Issue #5, opened 2018-12-22
LayerCake and SSR?


zzolo:
I haven't tried to do it, but it seems like it wouldn't work since LayerCake depends on a DOM element.  Maybe there is some way around this that I don't know about.

mhkeller:
Good question. I haven't played around with using the ssr mode but you're right that it does need that DOM element to measure stuff. Is there a way to know if you're in a server environment? In that case you could accept a default width and height.

I put in the ability to [pass any options](https://github.com/mhkeller/layercake/blob/a5f2ccc25f8d2dd554731f9272a242a4b8a17a35/src/layercake.js#L244) to the `.render` method so you can do `hydrate: true` and that should behave as normal. Let me poke around a bit and also look to what Svelte 3 would do with ssr.

mhkeller:
And if you've done any ssr setups and have an idea let me know!

zzolo:
For reference: [Svelte 3 store proposal](https://github.com/sveltejs/rfcs/blob/7e46f10fa93203066fa0dbd892f44a86ecd220b9/text/0002-reactive-stores.md).

Yeah, I think a component-based approach would be great.  I don't think it would really be all that different, outside usage of course.

The main question, imo, is how to pass the top-level properties around.  It could be done with component properties, or even could be a store as well, though neither is great.  It does sound like this part would be pretty straight-forward in Svelte 3.

mhkeller:
Now with 2.0.0, you get some SSR rendering for free but it's not all the way. Rich has a project that's more thought out support for javascript-less, responsive charts that I think will have some lessons to pull from. But give the new version a whirl and see how it works.

mhkeller:
Here's Rich's SSR technique: https://dev.to/richharris/a-new-technique-for-making-responsive-javascript-free-charts-gmp

The only thing that would have to change to do these kinds of charts with LayerCake is to make a separate Svg layout export that exposes the viewBox settings add `preserveAspectRatio="none"` and adds the non-scaling-stroke CSS which very easy to do.

```css
svg * {
  vector-effect: non-scaling-stroke;
}
```

 It could be called `SvgPercent` or something `SvgScalable`, `SvgPancake` as a reference to Rich's tool. The user would set a manual x and y range of `[0, 100]` so that the scales output in percentages. 

#### Separately

[...]

PR #7, opened 2018-12-26
SSR experiments

zzolo:
Here is an experiment with SSR.  This pull request is not actually meant to be merged, as it may not be the best approach.

Overall, there are a couple things that hold LayerCake back from SSR:

* DOM API parts.  Particularly, getting the width and height and padding from the element.
* Rendering, specifically `oncreate`, doesn't work as it doesn't get called in SSR.

This approach uses a `ssr` option in the config.  This could maybe handled automatically by checking for `window` or `document`.  The changes to the library are simply avoiding the DOM API.

To get it then to render out HTML, instead of calling `render`, this approach uses the `LayerCakeContainer.html` directly in the Svelte template.

There's an example included, where `examples/svelte-ssr.html/App.html` is the main template and what we run through Svelte's compiling.  The `example/svelte-ssr.js` is the small script to do the rendering.  So, you can do something like `node examples/svelte-ssr.js > examples/svelte-ssr.output.html` to create an HTML file that can be loaded in the browser.

### DIFF
--- a/src/layercake.js
+++ b/src/layercake.js
@@ -29,17 +29,20 @@ export { calcExtents };
 export default class LayerCakeStore extends Store {
 	constructor (config) {
 		/* --------------------------------------------
-		 * Set border box so padding works correctly
+		 * Set border box so padding works correctly. In SSR mode,
+		 * there's no need to set this.
 		 */
-		config.target.style['box-sizing'] = 'border-box';
+		if (!config.ssr) {
+			config.target.style['box-sizing'] = 'border-box';
+		}
 
 		/* --------------------------------------------
 		 * Main values
 		 */
 		const coreValues = {
 			data: config.data,
-			containerWidth: config.target.clientWidth,
-			containerHeight: config.target.clientHeight,
+			containerWidth: config.width || config.target.clientWidth,
+			containerHeight: config.height || config.target.clientHeight,
 			layouts: [],
 			target: config.target,
 			custom: config.custom || {}
@@ -121,10 +124,16 @@ export default class LayerCakeStore extends Store {
 	}
 
 	computeValues (settings, originalSettings) {
-		this.compute('padding', ['target', 'containerWidth', 'containerHeight'], target => {
+		this.compute('padding', ['target', 'containerWidth', 'containerHeight', 'ssr'], (target, containerWidth, containerHeight, ssr) => {
 			const defaultPadding = {top: 0, right: 0, bottom: 0, left: 0};
 			let hasPadding = false;
 			const padding = {};
+
+			if (ssr) {
+				// TODO: Can padding be defined in config if it's a computed
+				// value?
+				return defaultPadding;
+			}
 
 			const styles = window.getComputedStyle(target);
 			Object.keys(defaultPadding).forEach(p => {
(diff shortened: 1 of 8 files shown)


src/layercake.js
```diff
@@ -121,11 +124,17 @@ export default class LayerCakeStore extends Store {
...
 
 	computeValues (settings, originalSettings) {
-		this.compute('padding', ['target', 'containerWidth', 'containerHeight'], target => {
+		this.compute('padding', ['target', 'containerWidth', 'containerHeight', 'ssr'], (target, containerWidth, containerHeight, ssr) => {
 			const defaultPadding = {top: 0, right: 0, bottom: 0, left: 0};
 			let hasPadding = false;
 			const padding = {};
 
+			if (ssr) {
+				// TODO: Can padding be defined in config if it's a computed
```
mhkeller: You should be able to return the same `Object.assign(defaultPadding, settings.padding || {});` no?

mhkeller:
Thanks for putting this together! I think this makes a lot of sense and is pretty simple. I left a comment on the padding — I think you can return the same thing in ssr mode as in the default "no-css padding" mode. 

How would it work if you wanted it to stay responsive? Looking over the svelte docs I don't see how to get the javascript when registering the svelte ssr module. Maybe one has to use `.compile` for that to work?

mhkeller:
Aside to keep in mind: Layer Cake will have to change a bit to work with Svelte 3. The biggest difference is there is no more concept of a "Store". Obviously a larger todo but something to keep in mind.

src/layercake.js
zzolo: That makes sense to me.

zzolo:
Re: responsive.  Once the client side is loaded, the responsive stuff will kick in.  This is acceptable to me; my overall goal, is to have something there on load, then have it enhance once the client code is loaded.

If we are assuming SVG (which SSR will only be applicable to HTML and SVG anyway), there is the `preserveAspectRatio` to make the SVG responsive with just "styling".  This might breakdown the logic around scales, specifically for HTML and other containers.  And this responsive approach for SVG is usually not good enough anyway, since generally content needs to change in a responsive chart.

So, I think the big question is if there should be a `ssr` config option, or if you want to do a check for `window` and just go from there.  Otherwise, it's actually a pretty simple update to `src/layercake.js` plus some more significant documentation updates.

mhkeller:
> So, I think the big question is if there should be a ssr config option, or if you want to do a check for window and just go from there. Otherwise, it's actually a pretty simple update to src/layercake.js plus some more significant documentation updates.

Setting `target` or `width` + `height` are mutually exclusive, right? So if you're setting a width and a height, it can be assumed you're in SSR mode. If you set a target, you have access to the DOM and in the browser? I played around a little bit last week trying to get the client side portion up but failed. It was a bummer. If you have an end-to-end example where it hydrates the baked out SSR one that would be rad.

> This approach where the LayerCake store is defined in the component means that it doesn't have access to any of the components properties. This causes an issue if I am assigning the "data" to that component.

What's the full example look like? If you can send a test repo I'll take a look.

zzolo:
Ok, made an example repo.  But, ultimately, this isn't really a pull request yet, so moving discussion back to the issue #5 .

Issue #46, opened 2021-06-11
Replace quadratic-time algorithm for uniques()?


jtrim-ons:
The uniques() function checks if each item is in an array of seen items, which I think leads to quadratic complexity overall. There are probably use cases such as a scatter plot with thousands of points where this would take a noticeable length of time.

Would it be worth making `seen` a `Set`? I think it would just need the following changes:

- `const seen = [];` -> `const seen = new Set();`
- `seen.includes(computed)` -> `seen.has(computed)`
- `seen.push(computed)` -> `seen.add(computed)`

mhkeller:
Ah very good idea! This function was a holdover from another era.

jtrim-ons:
Great - I'll have a go at a PR just now.

mhkeller:
Thanks for jumping in! I'll take a look!

mhkeller:
Sorry for the delay! This is now published as v4.1.1! Thanks again! I'll hopefully get some time in the coming two weeks to go over the calcExtents API change, too.

PR #47, opened 2021-06-11
Use a Set to store seen values in unique(), for speed

jtrim-ons:
This replaces the `seen` array with a `Set`; see #46.


mhkeller:
Thanks @jtrim-ons!

PR #52, opened 2021-12-08
Add `"type": "module"` to `package.json`

benmccann:
The latest version of Vite will log a warning about this library being incorrectly packaged because the entry point is ESM without specifying either `"type": "module"` or having an `.mjs` extension. This will fix the issue

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,8 @@
     "test": "mocha --require esm test/*.js"
   },
   "license": "MIT",
-  "main": "src/index.js",
+  "type": "module",
+	"main": "src/index.js",
   "module": "src/index.js",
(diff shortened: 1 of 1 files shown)


mhkeller:
Thanks @benmccann. I haven't caught up on all the packaging requirements these days – was waiting to do that when I convert this lib to using svelte-kit. Does this PR have any implications for existing users? I made the change locally and it seems to still work using the sapper setup I have for the website (svelte kit migration is in progress there.)

benmccann:
I don't think it should affect end users other than getting rid of the console warning

mhkeller:
Thanks! Published as 5.0.2.

PR #55, opened 2021-12-15
Add svelte to package.json.  Fixes #54

### LINKED ISSUE #54: SvelteKit 1.0.0-next.202 / Vite 2.7 issue
After upgrading to SvelteKit `1.0.0-next.202` which [includes Vite 2.7](https://github.com/sveltejs/kit/blob/master/packages/kit/CHANGELOG.md#100-next202), I receive the following when importing LayerCake:

```
10:36:36 PM [vite] Error when evaluating SSR module /src/lib/components/Chart.svelte:
TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension ".svelte" for /Users/techniq/Documents/Development/open-source/layerchart/node_modules/layercake/src/LayerCake.svelte
    at new NodeError (node:internal/errors:371:5)
    at Object.file: (node:internal/modules/esm/get_format:72:15)
    at defaultGetFormat (node:internal/modules/esm/get_format:85:38)
    at defaultLoad (node:internal/modules/esm/load:13:42)
    at ESMLoader.load (node:internal/modules/esm/loader:303:26)
    at ESMLoader.moduleProvider (node:internal/modules/esm/loader:230:58)
    at new ModuleJob (node:internal/modules/esm/module_job:63:26)
    at ESMLoader.getModuleJob (node:internal/modules/esm/loader:244:11)
    at async ModuleWrap.<anonymous> (node:internal/modules/esm/module_job:78:21)
    at async Promise.all (index 0)
```

While I'm not 100% certain, I believe the heuristic changed with Vite 2.7 to detect a svelte package, and the best tell now is to include a `svelte` property in `package.json` (along with `main`/`exports`).  If using SvelteKit to build packages, running `svelte-kit package` [now](https://github.com/sveltejs/kit/pull/2431) sets this for you.  This PR along with the associated [issue](https://github.com/sveltejs/kit/issues/1959) and [code](https://github.com/sveltejs/kit/blob/master/packages/kit/src/packaging/index.js#L120-L145) provides some additional info.

Sending a PR your way...

techniq:


mhkeller:
Thanks @techniq !

mhkeller:
Published as 5.0.3. Let me know how it goes.

techniq:
@mhkeller Thanks for the quick release.  Error is gone :)

mhkeller:
Great. I was doing some work on the svelte-kit version. I just pushed up a version. If you want to give it a try that would be great: https://github.com/mhkeller/layercake/tree/kit

In particular I want to see if the typescript hints work.

techniq:
@mhkeller I spoke too soon... out of no where (on a hot reload) I started getting `Missing "./package.json" export in "d3-scale" package` errors on startup and now for the life of me can't work around it.

I'm fairly certain it's because none of the d3 packages do in fact have the `package.json` included in their [exports](https://github.com/d3/d3-scale/blob/main/package.json#L30-L33) but then I'm also not sure I'm up for submitting a PR for the ~50+ d3 repositories and how long they would take to get merged / released.

I setup a simple StackBlitz [example](https://stackblitz.com/edit/sveltekit-gw2gco?file=package.json) and going to log an issue in the SvelteKit issues to see if they have any thoughts.

What's odd is I couldn't reproduce it if I import `LayerCake` or `d3-scale` directly, but working directly with `LayerChart` or importing it gives these errors.

I'll let you know what I find out.

mhkeller:
I saw that error in a quick attempt i did at `npm link`ing the layercake.graphics repo to the `package` dir in the new svelte-kit layercake branch. but i wasn't sure what to make of it. one idea is that i thought it was related to better defining exports. currently, the kit branch exports all of the functions in `src/lib` including the helpers and utils, which previously were only internal functions. i wonder if it's trying also to make d3-scale an export? not sure.

another idea is to try to package the `scaleLinear` and `scaleSqrt` functions inside of layercake and remove that dependency – not sure how feasible that is. it also means i would want to regularly publish new versions whenever those dependencies get patches.

mhkeller:
@techniq let's move this discussion to this issue: https://github.com/mhkeller/layercake/issues/56

Issue #71, opened 2022-04-22
Svg and Canvas attribute control


maxblee:
Hello, I was wondering if it might be possible to add attribute-level controls on the `<svg>` and `<canvas>` elements (so `Svg`, `ScaledSvg`, `Canvas`, and `Webgl`). My most common use-case for this comes in wanting to add descriptions for svgs. In order to do this in HTML, I would write something along the lines of

```html
<svg role="img" aria-labelledby="my-label">
    <title id="my-label">This is a label.</title>
</svg>
```

Currently,  I should be able to add the `role` and `aria-labelledby` attributes on the `Svg` component by running `setAttribute()` on the element bindings, but that feels a little bit clunky, especially for something I use fairly regularly. I'd prefer being able to write something along the lines of

```svelte
<Svg role="img" aria-labelledby="my-label">...</Svg>
```

or

```svelte
<Svg extraAttrs={{role: "img", "aria-labelledby": "my-label"}}>...</Svg>
```

If you'd be interested in this, let me know; I'd be happy to submit a pull request. Thanks!

mhkeller:
Hey @maxblee Yea good idea. I like the second approach since it covers future use cases. And probably just `attrs` as the prop name works. Could you start a PR? Would this not apply to the HTML layout component too? In your PR, if you could also add info in the docs with examples that would be great. It would go in [this file](https://github.com/mhkeller/layercake/blob/main/src/content/guide/05-layout-component-props.md).

maxblee:
@mhkeller Sorry, I meant to respond/get to this sooner but got lost in other things. I've had a chance to take an initial stab at this. Similar to you, my first thought was to use spread props. But for some reason, they seem to create some sort of conflict with [the scaleCanvas utility](https://github.com/mhkeller/layercake/blob/main/src/lib/lib/scaleCanvas.js). I saw this when I built the static SvelteKit site and viewed [the Voronoi example](https://layercake.graphics/example/Scatter/). When I didn't include the spread attributes `{...attrs}`, `scaleCanvas` worked properly and created a canvas with a fixed width and height, like

```html
<canvas class="layercake-layout-canvas" style="width: 770px; height: 170px; position: absolute; inset: 10px 5px 20px 25px;" width="1540" height="340"></canvas>
```

But with the spread attributes, the width and height stayed at 100%, even though `scaleCanvas` seemed to run .

```html
[...]

PR #74, opened 2022-05-16
Draft: add attribute control for components + docs

maxblee:
This adds attribute-level control to the `Svg`, `ScaledSvg`, `Canvas`, `Html`, and `Webgl` layout layers. Addresses #71 .

### DIFF
--- a/src/lib/layouts/Canvas.svelte
+++ b/src/lib/layouts/Canvas.svelte
@@ -19,6 +19,19 @@
 	/** @type {Boolean} [pointerEvents] Set this to `false` to set `pointer-events: none;` on the entire layer. */
 	export let pointerEvents = undefined;
 
+	/** @type {Object|null} [attrs] An object that sets additional attribute values onto the `<canvas>` tag*/
+		export let attrs = null;
+
+	$: if (element && attrs) {
+		const attrList = Object.entries(attrs);
+		for (let i = 0; i < attrList.length; i++) {
+			const [attr, property] = attrList[i];
+			if (!element.hasAttribute(attr) || element.getAttribute(attr) !== property) {
+				element.setAttribute(attr, property);
+			}
+		}
+	}
+
 	const { width, height, padding } = getContext('LayerCake');
 
 	const cntxt = {
--- a/src/lib/layouts/Canvas.svelte
+++ b/src/lib/layouts/Canvas.svelte
@@ -5,7 +5,9 @@
 <script>
 	import { getContext, onMount, setContext } from 'svelte';
 	import { writable } from 'svelte/store';
+
 	import scaleCanvas from '../lib/scaleCanvas.js';
+	import setElementAttributes from '../utils/setElementAttributes.js';
 
 	/** @type {HTMLCanvasElement} [element] The `<canvas>` tag. Useful for bindings. */
 	export let element = undefined;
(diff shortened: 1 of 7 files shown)


mhkeller:
Thanks again for this. I looked up a [performance benchmark](https://www.measurethat.net/Benchmarks/Show/3686/0/object-values-objectentries-vs-objectkeys-vs-objectkeys#latest_results_block) and it seems that `Object.keys` is slightly faster than `Object.entries`. I switched it but do you have a test scenario for this? If not, we should add this functionality to one example for each layout so we can use the website as a test.

src/lib/layouts/Canvas.svelte
```diff
@@ -19,6 +19,20 @@
...
 
+	/** @type {Object|null} [attrs] An object that sets additional attribute values onto the `<canvas>` tag*/
+		export let attrs = null;
+
+	$: if (element && attrs) {
+		const attrKeys = Object.keys(attrs);
+		for (let i = 0; i < attrKeys.length; i++) {
+			const attr = attrKeys[i];
+			const property = attrs[attr];
+			if (!element.hasAttribute(attr) || element.getAttribute(attr) !== property) {
```
mhkeller, line 30: Let's say you wanted to dynamically change the label post-render, would this conditional prevent that because the `element.hasAttribute(attr)` won't set any attribute that is already set?

mhkeller (changes requested):

src/lib/layouts/Canvas.svelte
mhkeller: Probably getting rid of `!element.hasAttribute(attr) || ` would be good.

mhkeller:
I moved this into its own function so it wasn't repeated and I also added some type checking and a warning message. One question though is what if you want to dynamically unset these properties? Should this function make sure that the properties are always synced with the value of `attrs`? That seems most sveltey and similar to using spread props.

mhkeller:
One idea is to prefix these attributes with something and every time this function runs, it removes prefixed attributes that are not present in `attrs`

mhkeller (approved):

maxblee:
Sorry for taking so long to get back on this. I definitely think it makes sense to be able to dynamically unset the properties. I imagine you could use an internal `Set` to keep track of which attributes were initially set, then compute the difference to figure out which elements, if any, were unset. But there might be a more elegant way to handle it as well.

I've checked locally to make sure everything built properly but agree that we should add a test scenario, particularly one that handles unsetting/resetting attributes.

Also, a quick note on the `hasAttribute` call that I made in the initial pull request: The reason I put that in there was to handle boolean attributes like `required` or `hidden`. According to the DOM spec, `getAttribute` [should return an empty string on non-existing attributes](https://developer.mozilla.org/en-US/docs/Web/API/Element/getAttribute#non-existing_attributes), which would also be what you'd expect to see from an existing boolean attribute. However, it looks like modern browsers return `null` instead of adhering to the DOM spec, so maybe not an actual concern...

mhkeller:
I don't totally understand. How does the `hasAttribute` check fix / handle / help this scenario? And is there a way to solve this scenario in a way that allows for changing an existing property?

maxblee:
Sorry I wasn't clear. Say you want to pass this object as `attrs` in the `setElementAttributes` function:

```js
const attrs = { hidden: "" }
```

If you did this, according to the DOM spec, `element.getAttribute('hidden') !== attrs['hidden']` would return `false` since `getAttribute` and `attrs['hidden']` would both return empty strings. However, `element.hasAttribute('hidden')` would return false. So by replacing the check with `element.getAttribute(attr) !== property || !element.hasAttribute(attr)`, `element.setAttribute(attr, property);` would run where it otherwise wouldn't (according to the DOM spec). Again, modern browsers seem to all return `null` on non-existing properties, so this is probably a moot point.

If I'm not mistaken, this should already allow for changing an existing property, since `element.getAttribute(attr) !== property` will return false if someone changes an attribute from its initial value. The tricky thing is going to be handling attributes that are removed. I think you could handle that case by holding an internal `Set` that keeps track of the attributes that have been set on the element, computing the difference between that set and `Object.keys(attrs)`, and either running `element.removeAttribute(attr)` or resetting the attribute to its original value (if someone changed one of layercake's computed attributes). But there might be/probably is a more elegant + performant way to handle that, and agreed that it makes sense to see what Svelte does with spread props and $$restProps.

mhkeller:
I'm going to close this in favor of https://github.com/mhkeller/layercake/pull/133 since that will add the ability for aria-labels without the possible performance issues cited here. Let me know if there's a good reason for continuing with this approach, though, or if there are other commonly used props that we should pass to layout components.

Issue #61, opened 2022-01-19
Make some map examples using pre-projected data


mhkeller:

mhkeller:
Closed from https://github.com/mhkeller/layercake/pull/77 Thanks @chris-creditdesign

PR #77, opened 2022-05-18
Adds map example using pre-projected data ref #61

chris-creditdesign:
- Adds MapSvgPreProjected components in 'examples' and '_examples_ssr' folders.
- Adds links to editable versions on the svelte repl.
- Adds pre-projected toposjon file.


mhkeller:
Thanks @chris-creditdesign! I should be able to review this in the next week or two.

mhkeller:
Thanks a bunch for this. I didn't realize that what you do is pass in `geoIdentity` as the projection. I think in order to keep the home page consistent in the two-charts-per-row style. I may replace the existing map example with this one!

PR #82, opened 2022-06-24
Fix typos with affected ranges

techniq:
Found a few typos while reading the docs :)


mhkeller:
Thanks!

mhkeller:
Published as v6.1.1

Issue #83, opened 2022-06-24
Consider not reversing yScale by default for scaleBand


techniq:
[Currently](https://github.com/mhkeller/layercake/blob/main/src/lib/settings/defaultReverses.js#L3) the yRange is always [reversed](https://layercake.graphics/guide#yreverse), but this can lead to some unexpected results when using `scaleBand` as the defined `yDomain` order not being maintained (rendered bottom up instead of top down).  For example:

- https://layercake.graphics/example/Bar/
- https://layercake.graphics/example/BarStacked/
- https://layercake.graphics/example/ClevelandDotPlot

You can work around this two different ways:
- Reverse the domain passed, as done in this example: https://layercake.graphics/example/Timeplot/
- Set `yReverse={false}`

The former is what I've been doing (mostly out of trial/error), although it has caused me other issues when things expect a certain order, like bisecting or [inverting a scaleBand](https://github.com/techniq/layerchart/blob/master/src/lib/utils/scales.ts#L5-L23).

While there is typically not a great way to identify the type of scale, luckily scaleBand (and scalePoint) can be identified by checking the present of `scale.bandwidth` which I do in [LayerChart](https://github.com/techniq/layerchart/blob/master/src/lib/utils/scales.ts#L27-L29) quite often.

I might change how I handle `yReverse` in LayerChart to accommodate this, but wanted to run it by you first, as I try not to deviate too much :).  I do have a [Chart](https://github.com/techniq/layerchart/blob/master/src/lib/components/Chart.svelte) component that provides a few conveniences , some of which might be worth moving up to LayerCake.

mhkeller:
Hey thanks for this. Would using the yReverse option fix those secondary issues you’re seeing when reversing the order. 

This is for sure a tough one. Here’s a summary of the thinking that led to this: I decided to make the reverse be the default since it makes getting up and running much easier. From there, it seemed good to just keep things consistent and not have dynamic properties, I.e. the value of yReverse defaulting to false for bandwidth scales.

However, that could be a consideration more in the abstract sense of the library and users would generally prefer to just have as many chart types work out of the box as possible. 

Curious to hear any other thoughts but this seems like a reasonable change. Are there any scenarios where the changing nature of yReverse is problematic?

techniq:
[...]

PR #85, opened 2022-06-28
Set yReverse={true} unless yScale={scaleBand()}.  Issue #83

techniq:
Created a quick PR based on discussion in #83.  A few notes:

- I removed `defaultReverses.js` instead of changing it to a function that took in the applicable scale, as it seemed unnecessary.
- Moved `yReverse` (et. al.) below `yScale` (et. al.) since it's now dependent
- There is a typescript error regarding `yScale.bandwidth` since it's just typed as a `Function`.  I've always struggled with dealing with d3-scales in typescript cleanly, so not sure there is much improvement here (beyond bandaids).
- Using `max($yRange)` instead of `$yRange[0]` in `AxisX` since the max is now conditionally the start/end of the domain.  I just did something similar throughout [LayerChart](https://github.com/techniq/layerchart/issues/29)
- Removed explicit `.reverse()` in time of day examples since no longer needed
- All other examples mentioned in #83 (Bar, BarStacked, ClevelandDotPlot) will now be inverted from how they used to be, but I think this is the correct order (matches order of `yDomain` provided) and is basically the intent of the PR.
- I tried to update the jsdoc / guide, but wasn't 100% sure how much was generated instead of being manually updated, so I've likely overlooked something.

I didn't do an exhaustive check for breaking changes, but spot checked 6 or 7 examples and didn't find any regressions.

Feel free to take this PR and modify as you like.  I tried to always save without formatting so my prettier styles would dirty the diff :).


mhkeller:
Thanks! I'll go over this in the coming weeks. I have to go through https://github.com/mhkeller/layercake/pull/74 and merge that first since #83 will be a breaking change so good to push some other stuff in first

src/_components/AxisX.svelte
```diff
@@ -54,7 +55,7 @@
 
 <g class='axis x-axis' class:snapTicks>
 	{#each tickVals as tick, i}
-		<g class='tick tick-{ i }' transform='translate({$xScale(tick)},{$yRange[0]})'>
+		<g class='tick tick-{ i }' transform='translate({$xScale(tick)},{max($yRange)})'>
```
mhkeller, line 58: Could you do `Math.max(...$yRange)` instead to avoid the `d3-array` import?

techniq: I can.  I only used `d3-array` since you already had it as a `devDependency` and it was a little cleaner (and what I used in LayerChart).

I can switch to `Math.max(...$yRange)` if you prefer.

mhkeller: Yea may as well just to make it more vanilla where possible

mhkeller:
Fyi, this also sets `yReverse=false` if the scale is a `scalePoint`. I think that's the desired behavior, just wanted to mark that it applied to a scale other than `scaleBand`.

techniq:
Thanks. I saw that, and to be honest, never had the need for a scalePoint,
but assumed it worked the same :)

mhkeller:
Yea i think it's new? i think timeplot or clevelanddotplot could be a candidate for it...

Issue #27, opened 2020-06-17
Dynamically calculate unique domain entries for .bandwidth scales


mhkeller:
Currently, Layer Cake measures the extent of the data and the dimensions of the div to create scales. This is most useful to linear scales and all others require you to calculate your own domain, such as for `scaleOrdinal`. It would be nice if the user put in a `scaleOrdinal`, Layer Cake took the key for that dimension and calculated unique fields on its own. The user could override these just as they can do currently but it would be a nice way to get quickly up and running. 

The blocker on this is that there's no easy way to determine what d3 scale is being used since they're all duck typed. You could write out logic to figure out what the scale is based on the methods but that seems easy to mess up and hard to maintain and test. You could pass a string for the scale but then you'd have to import all of d3-scale – not sure if that would be tree-shook. This is on hold for now but it's an interesting idea if other progress can be made.

mhkeller:
I'm going to close this for now since it seems unlikely d3-scale will add the necessary properties to make this workable.

mhkeller:
I think you could do this just for scales that have a `.bandwidth` method and it would be fairly easy.

mhkeller:
domain props could also take a function that returns the dynamically calculated domain so you could do a custom sort or something.

mhkeller:
Should this be sorted by default?

mhkeller:
I'm leaning towards not having it sort automatically. it's easy to do yourself but a default sort can't be undone

techniq:
@mhkeller I don't think they should be sorted directly... I figure to generate the `domain` we would just `map` and `unique` the results (maintaining the original order).  If you wanted a specific sort, you would just presort the data (if possible based on desire), else you could just pass the `xDomain`/etc directly and sort it as desired.

One complicated example (where I think it's best to let the user either sort the `data` or pass in the `xDomain` explicitly is when you have a timeline with multiple durations within a band.  You might want to sort the bands(items) by the first starting time, last/most recent starting time, or based on ending time.

Something like this...

<img width="1189" alt="image" src="https://user-images.githubusercontent.com/177476/178165785-02b9af0f-7cc7-4a4f-911f-3affa7216633.png">

I've had this use case for a page view chart where I either sort by the first time a user visited the site, or the most recent.

mhkeller:
Good points! Thanks for weighing in. Not sorting it I think is the best option. There are many alternatives for having the sort option.

mhkeller:
This is added in v7.0

PR #90, opened 2022-07-09
Calculate unique values for ordinal scales

mhkeller:
Solves https://github.com/mhkeller/layercake/issues/27


mhkeller:
The only weird thing in this is I modified the `flatten` function to pass string keys on arrays onto the child object. This could cause issues such as key globbering or other object distortions and I may not include it. The reasoning for this for the moment is there is no other way to detect series names when using a wide data structure in stack layouts. This way, it copies the `key` field from the stack group array onto each child array, which is what is passed to the `calcUniques` function with `flatData` as the argument.

mhkeller:
I got rid of the weird stuff in `flatten`. In all of those stacked examples, you compute the domain list manually anyway so there's no extra work except setting it as a prop. if the user wanted to, they could loop through the stacked data and add their own value on the z key

Issue #93, opened 2022-07-22
Add transform components


mhkeller:
I was thinking it could be useful to provide for various transform buckets that wrap `<LayerCake>` components and whose job it is to transform your data and then pass it back out. Some ideas

1. **Stack** To stack your data. It would return variables for `data` and `flatData`
2. **Group** For things like converting from wide to long data and grouping. Maybe `WideToLong` is a separate component
3. **Facet** This was the main use case that got me thinking. Similar to group but it creates small multiples. First attempt: https://github.com/mhkeller/layercake/blob/transforms/src/lib/transforms/Facet.svelte
5. **Bin** For easier histogramming

These should be able to be combined such as `Facet` and `Bin` could created a faceted histogram.

mhkeller:
I'm adding some of these as helper functions but removing the facet component because it is impractical to make as a general purpose component in the library.

PR #95, opened 2022-07-31
Draft: Add transform components

mhkeller:
Per https://github.com/mhkeller/layercake/issues/93


mhkeller:
This is mostly done. I need to harmonize some of the examples so that they use this and then also some don't use them so people can see it both ways. It may be most helpful to have a new section on the website, too, similar to the component gallery.

mhkeller:
Not sure if these are really all that helpful...

mhkeller:
I kept the Facet component but converted the others to simple helper functions

mhkeller:
Actually the facet one doesn't make for a good general use component, even if it is a fun idea

mhkeller:
Here's a fun example though for posterity: https://svelte.dev/repl/f71eebb8f0014bbe93739937a122bbe4?version=3.55.0

PR #101, opened 2022-10-02
Support passing `count` as argument to `xNice` prop (et al.)

### LINKED ISSUE #99: Pass count to `nice`
It would be helpful if you could pass the tick [count](https://github.com/d3/d3-scale#continuous_nice) to [yNice](https://layercake.graphics/guide#xnice) (et. al) instead of only a boolean and it would get set in [createScale()](https://github.com/mhkeller/layercake/blob/main/src/lib/helpers/createScale.js#L42).

So we would still support...
```svelte
<LayerCake yNice />
```
...which would call `.nice()`

but also support...
```svelte
<LayerCake yNice={4} />
```
...which would call `.nice(4)`

I created this REPL to [demonstrate](https://svelte.dev/repl/92e0fc6e7e6444c8b822c3af824cf085?version=3.50.1) the flexibility of passing the count to `nice`.

techniq:
Resolves #99

Currently...

image

...after passing `yNice={4}` (which matches [AxisY](https://github.com/mhkeller/layercake/blob/main/src/_components/AxisY.svelte#L20) `ticks = 4` default)

image

### DIFF
--- a/src/lib/helpers/createScale.js
+++ b/src/lib/helpers/createScale.js
@@ -39,11 +39,7 @@ export default function createScale (s) {
 
 		if ($nice === true || typeof $nice === 'number') {
 			if (typeof scale.nice === 'function') {
-				if (typeof $nice === 'number') {
-					scale.nice($nice);
-				} else {
-					scale.nice();
-				}
+				scale.nice(typeof $nice === 'number' ? $nice : undefined);
 			} else {
 				console.error(`[Layer Cake] You set \`${s}Nice: true\` but the ${s}Scale does not have a \`.nice\` method. Ignoring...`);
 			}
--- a/src/lib/helpers/createScale.js
+++ b/src/lib/helpers/createScale.js
@@ -37,9 +37,13 @@ export default function createScale (s) {
 			scale.domain(padScale(scale, $padding));
 		}
 
-		if ($nice === true) {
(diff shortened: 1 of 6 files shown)


techniq:
@mhkeller I'm never 100% sure which docs are generated and which need to be updated, so you might want to double check I didn't overlook something.

If this looks good and gets merged, could you publish as `7.1.0`?  Thanks.

src/lib/helpers/createScale.js
```diff
@@ -37,9 +37,13 @@ export default function createScale (s) {
 			scale.domain(padScale(scale, $padding));
 		}
 
-		if ($nice === true) {
+		if ($nice === true || typeof $nice === 'number') {
 			if (typeof scale.nice === 'function') {
-				scale.nice();
+				if (typeof $nice === 'number') {
+					scale.nice($nice);
```
mhkeller, line 43: Could put this as a conditional inside of the function call?

mhkeller (approved):

src/routes/_examples/Line.svelte
```diff
@@ -35,6 +35,7 @@
 		padding={{ right: 10, bottom: 20, left: 25 }}
 		x={xKey}
 		y={yKey}
+		yNice={4}
```
mhkeller, line 38: Gonna add this to the SSR example too.

Issue #103, opened 2022-11-19
Expose `LayerCake` context as slot props


sawyerclick:
**What**
Expose the `LayerCake` context calculations as slot props for the `<LayerCake>` component to prevent overcomplicated file structures for simple graphics.

**Why**
I often find myself wanting to compose single-file Cakes (an un-LayeredCake, if you will), but run into the issue of _needing_ the context's calculations and getters. If we were to expose the calculations as slot props, you'd be able to pull in repeated patterns, such as axes, while preventing the need to move every single calculation-dependent node into a component. 

One example is the multi-line chart, where I wish to iterate over my data to create individual `<Line>` components directly in a `<Svg>` slot, a feature-rich component already in my toolkit. 

**How**
Since stores must be declared at the top level of the component, we'd have to expose the regular variables. This would involve having to first create the variables as non-stores

```svelte
<slot {...regContexts}>
``` 

or

```svelte
<slot xGet={$xGet} yGet={$yGet} ...>
```

mhkeller:
Interesting idea. The simplest workaround would be to make one component as the direct child of the LayerCake component and put everything in that if you want to avoid multiple components.

I don’t fully get the first example with regVariables. Can you show more context. Your second example makes sense to me and I imagine doing it that way will preserve all the same variable reactivity? I wonder if there is a performance or size penalty for exposing more slot props. If you have any comparisons that would be great. 

I have a few other features and documentation changes on my todo list to finalize first.

mhkeller:
@sawyerclick just checking in on this since it seems like a good idea to implement if there's no performance hit.

sawyerclick:
howdy! thanks for checking in. day job got busy so I haven't had the time, but I can look into this now. I'll fork and apply changes, benchmarking the perf differences.

Additionally, after some thinking, I do think this approach is more appropriate. Though it does bring scalability into play if you ever added any more props/contexts.

```svelte
<slot xGet={$xGet} yGet={$yGet} ...>
```

mhkeller:
That sounds good. I don't mind the scalability issues – I don't foresee adding a huge amount of props in the future so having to write them all out once for this seems fine.

mhkeller:
And thanks for the time in testing out the changes!

Issue #56, opened 2021-12-15
Update to use svelte-kit (includes adding types)


mhkeller:
This is a work in progress on the [kit](https://github.com/mhkeller/layercake/tree/kit) branch. It includes typescript support and will close https://github.com/mhkeller/layercake/issues/49 

TODO

- [x] Add types for typescript
- [x] Figure out how to export fewer functions
- [x] Excludes .DS_Store
- [x] Get example working
- [ ] Add typescript types on context values (maybe slot values, too)?

mhkeller:
I added a basic example to the repo itself and it seems to work. Not sure why it wasn't working from pasting it into another project's node_modules but i could also see that breaking for a whole host of reasons...

mhkeller:
I added better export rules but still getting `Package subpath './package.json' is not defined by "exports" in`layercake/node_modules/d3-scale/package.json`

techniq:
@mhkeller Gosh that doesn't sound good, and I feel like I keep going deeper down a rabbit hole the more I follow.

mhkeller:
I would think that the vite setting of [optimizeDeps](https://vitejs.dev/config/#dep-optimization-options) would bundle the d3-scale options but that doesn't seem to have any effect.

as a fallback tho, layercake isn't doing anything fancy – it's just a series of components, really. so you can download the source into your project and import it like you would any regular component.

techniq:
In one of my private app repos I basically had all its npm packages referenced as it would help keep Vite from requiring multiple rescans on startup until it built its cache, and helped with SSR errors I was getting.  After upgrading SvelteKit/Vite 2.7, I was getting `TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension ".svelte" for ...` errors until I commented them out.  At the time, it appeared I could remove them all, but maybe I need to look into adding the d3/etc back.  I'm not sure how transitive deps will be though (LayerCake => d3-scale, for example) as I don't think i can add LayerCake back...

![image](https://user-images.githubusercontent.com/177476/146249233-0c4d5785-6735-4e77-93b5-74a5dd5ce03e.png)

Each major upgrade to SvelteKit/Vite always turns into hack a mole and then I'm afraid to touch things too much :(

mhkeller:
yea i'm very confused by it.

techniq:
[...]

Issue #49, opened 2021-11-16
Typescript types?


abalmos:
Sorry if this is the wrong place to ask, but don't see any other community forum/chat.

I just wanted to verify that I am not missing user contributed Typescript types for Layer Cake that you might be aware of?

mhkeller:
There aren’t but I should add them. Do you have an example repo that implements them?

abalmos:
I don't yet, but I have considered starting a set. Unfortunately, at the moment I don't have the free time to spare. Maybe soon -- thanks for confirming.

abalmos:
Are you open to converting the code base to TS? Looking at the source, I'm wondering if that isn't too much more work then creating a .d.ts file.

mhkeller:
Yeah I think it's a good idea – I just want to see how other svelte projects have implemented them / best practices etc.

mhkeller:
I'm tinkering with the best way to do this on the [kit](https://github.com/mhkeller/layercake/blob/kit/src/lib/LayerCake.svelte) branch. I'm finding a lack of great docs and examples, though so if anyone has any ideas on best practices, please let me know.

pierreminik:
I add my interfaces and types in the `src/global.s.ts` (without `export`) and then I don’t need to import them anywhere. I don’t know if this is best practise but it seems to work okay.

mhkeller:
The kit branch now has these if anyone wants to test: https://github.com/mhkeller/layercake/tree/kit

mhkeller:
I'm going to close these in favor of this issue: https://github.com/mhkeller/layercake/issues/56 any testing would still be appreciated!

abalmos:
@mhkeller Thanks for looking at this. I had to walk away from that project for a little while but I hope to get back to it by latest early next week. I will try out #56 and give feed/help as needed when I do so if the PR is still open.

In the meantime I had gone the same route as @pierreminik and built my own `types.d.ts` but your approach in #56 is much better long term.

PR #104, opened 2022-12-02
Add TypeScript support

allezxandre:
Here's my attempt at adding some type information to LayerCake, mainly the accessor attributes.  

Some notes:

- I've transformed some JS files to use the TypeScript extension. SvelteKit package compilation makes it quite easy to gradually adopt TypeScript, but I can understand if it's not something you want for LayerCake @mhkeller
- Having some files written in TypeScript means mochas tests won't work out of the box
- By making the `LayerCake.svelte` component use `script lang="ts"` with Generics, I was able to make it depend on the input data. This brought a number of advantages, namely with accessors that can now infer wether their output will be an array or not, but also letting the users know if their accessor becomes invalid
- Having the `LayerCake` depend on the typing of the input data means the Context now depends on both the input Data _and_ the accessors. It's a bit cumbersome to have to rewrite both types in every component that need to pass type info to `getContext`, so I found that by making them generic components that take `context` as a slot prop makes them easier to use. However that means I have to pass them to every component that wants to use accurate type data (a bit like what was suggested in #103, but still grouped in `context` rather than spread out like `{...context}`)

Now, usage becomes:

```svelte
<LayerCake x="x" y="y" {data} let:context>
  <Svg>
    <Component  {context} />
  </Svg>
</LayerCake>
```

Here, `x="x"` is type-checked and TypeScript will error out if `data` has a shape that is not compatible with the accessor.

And then inside the `Component` we can infer and constrain the type as a Generic like this:

image

Notice TypeScript knows that the `x` accessor returns a `number` and not a `number[]` because the `Context` has been contrained to extend `LayerCakeConfigSimple<DataRow>`. This contraint will also be checked in the caller component, where the `<Component  {context} />` tag will make sure the context provided by the parent satisfies the constraint.

There is still some work left, but as the move might become heavier than I expected I wanted to get some of your thoughts @mhkeller before I move forward.

Relates to #49 and #56.


mhkeller:
Hey! Thanks a bunch for putting this together and sorry for not getting to it yet. As a general note, I am not avid TypeScript user but I do like helpful popups as much as possible. So if there are ways to tune up LayerCake under the hood that allows for that while preserving the existing API as much as possible, that would be great. One question I have is on the context change, is that something that is required for all users or just if you want to use types?

allezxandre:
> One question I have is on the context change, is that something that is required for all users or just if you want to use types?

This is not required for all users, but those that want infered types will prefer it. Otherwise it's still possible to keep using getContext, even with types (but they will have to be input manually)

However after some more work on the typings, I think it will be very hard to have accurate typing everywhere. I thought that continuous scales matched the same interface but I stumbled upon the D3 Time Scale that doesn't match it. Because of that I would need to know the type of all accessors, the type of the input data, and now the type of each scale as well just to infer the context, and so the rabbit hole gets deeper. 

I'm afraid it's much more complex than I anticipated 😕

mhkeller:
Got it. Yeah it is quite a bit of work for a library that is mostly a pass-through and doesn't really care about what the signatures are of each of the scales. I'll go ahead and close this but maybe it can be inspiration for someone else! Thanks for the time in exploring it. If you see any incremental changes to the current setup, please feel free to open another PR.

Issue #111, opened 2023-02-07
Add debug mode to inspect scales and bounding box


mhkeller:
This feature makes it easy to check whether the scales picked up the right domain and range. You add `debug={true}` to the component and it prints out info about the container and all active scales. Would any other things be helpful to add?

**Here's a preview of the console output**

![Screenshot 2023-02-07 at 1 12 13 PM](https://user-images.githubusercontent.com/498744/217331592-10fd7c1d-d269-4396-ad9f-8e3b0ab4af8c.png)
![Screenshot 2023-02-07 at 1 10 50 PM](https://user-images.githubusercontent.com/498744/217331593-40e77c90-4b73-4ab3-87d3-e736d950459d.png)

mhkeller:
It would be cool if hex codes appeared as their color

PR #112, opened 2023-02-07
feat: debug mode

mhkeller:
Closes https://github.com/mhkeller/layercake/issues/111


mhkeller:
This is pretty much done except I haven't found a good way to differentiate `scaleUtc` from `scaleTime`. I need to either add a note to the output or [if this PR](https://github.com/d3/d3-scale/pull/274) is merged, we can more reliably use the name, which would be better in general.

mhkeller:
This is all set. I'm just waiting on https://github.com/mhkeller/layercake/pull/118 to be able to package the library again.

PR #118, opened 2023-03-12
Svelte package migration

mhkeller:
This updates the `npm run package` process to use the new svelte-package library. It defines `exports` and `typesVersions`.


mhkeller:
Tagging some Layer Cake typescript users to see if this `package.json` is set up correctly. If you could take a look and see if this is the easiest way to set up the exports and types, I would appreciate it: @abalmos, @techniq, @allezxandre 

I am using the `typesVersion` option as described [here](https://kit.svelte.dev/docs/packaging#typescript) but perhaps a better way to do it is to add documentation saying people need to:

> set the `moduleResolution` option in their `tsconfig/jsconfig.json` to `bundler` (available since TypeScript 5, the best and recommended option in the future), `node16` or `nodenext`.

Let me know if you have thoughts on what is currently easiest for typescript users.

allezxandre:
Sorry, I can't be of much help here, I've always used the default settings with TypeScript code, so I wouldn't know how integrating your own type definitions would work 😕

PR #140, opened 2023-07-13
Multiline show example data

rgieseke:
I thought it might be helpful to see the transformation of data formats in the MultiLine example by adding examples of the objects. 
Let me know if you want to include this (it does make it a bit longer to scroll before one sees the code) , then I can do the same for the SSR example and maybe look into formatting (with Highlightjs or just some improved styling).

It's the output of adding in `+page.svelte` of the example code (which I do every time I need a MultiLine plot).
```Js
$: console.log(data)
$: console.log(JSON.stringify(dataLong, null, 4));
$: console.log(flatten(dataLong, 'values'))
```  

It looks like this now:

image
image


mhkeller:
Great idea. Data formats and transforms are tricker so anything to help explain them is very good. Yeah see if you can get some highlighting going on there. Thanks for this!

mhkeller:
Great idea. Data formats and transforms are tricker so anything to help explain them is very good. Yeah see if you can get some highlighting going on there. Thanks for this!

rgieseke:
Just made the `code` block cover full width and adjusted the SSR example as well:

image

rgieseke:
I didn't really try to wrap my head around how the Markdown rendering and highlighting would interact as this looked good enough to me. :slightly_smiling_face:

Issue #117, opened 2023-03-05
types for xDomain


MCMattia:
Hi guys, I have been using your library to make a simple 3 line graph. It works very good. Thank you for your work.

I'm using yDomain in the following way:
`yDomain={[0, null]}`

Since the types for yDomain are:
`[min: ?Number, max: ?Number]`
Typescript is warning that `null` is not assignable to a number.

I think that the types for xDomain or yDomain, should be:
`Array:[min: ?Number|null, max: ?Number|null]|String[]|Number[]|Function`
instead of:
`Array:[min: ?Number, max: ?Number]|String[]|Number[]|Function`

Basically adding the null option since as you document on your website:
_If you want to inherit the value from the data's extent, set that value to null._

mhkeller:
Great catch – thanks! I'll put in a fix for this.

mhkeller:
@MCMattia this is done in [this branch](https://github.com/mhkeller/layercake/tree/svelte-package-migration). Svelte kit has updated the way that it packages libraries and it requires changing the deployment and packing system, which I won't be able to do today. I need to make sure it doesn't mess a bunch of things up. Hopefully it won't be too much trouble and you can add a ts ignore for now.

MCMattia:
Hi @mhkeller thank you for this work. For now I have manually fixed it in the `LayerCake.svelte.d.ts` file

mhkeller:
This is updated in 7.2.3

PR #141, opened 2023-07-17
Remove explicit null value from JSDOC type definition for domain values

rgieseke:
When using Layercake in a SvelteKit project with JSDoc or TypeScript type checking I am getting an error message for the `[0, null]` domain values, for example `yDomain` here:

```svelte
 <LayerCake
    padding={{ top: 7, right: 10, bottom: 20, left: 25 }}
    x={xKey}
    y={yKey}
    z={zKey}
    yDomain={[0, null]}
    zScale={scaleOrdinal()}
    zRange={seriesColors}
    flatData={flatten(groupedData, 'values')}
    data={groupedData}
  >
```

This gives the error message (from `npm run check`): 

```
Error: Type '[number, null]' is not assignable to type 'Function | number[] | string[] | [min: number, max: number] | undefined'.
  Type '[number, null]' is not assignable to type 'Function | number[] | string[] | [min: number, max: number]'.
    Type '[number, null]' is not assignable to type '[min: number, max: number]'.
      Type at position 1 in source is not compatible with type at position 1 in target.
        Type 'null' is not assignable to type 'number'. (js)
```        

When I change the JSDOC definitions to remove `|null' it still seems to work fine, so maybe there are not neccessary?

JSDoc.app lists nullable types like this as well

```
A number or null. {?number}
This indicates that the type is either the specified type, or null. 
```

https://jsdoc.app/tags-type.html


rgieseke:
I spoke too soon, this does not fix the problem :see_no_evil:

rgieseke:
Also didn't see the previous discussion in https://github.com/mhkeller/layercake/issues/117

mhkeller:
Hm – it would seem fairly basic to me that something could be null. Is it saying that it needs to have both types be the same? Do you get the same error if you set `[null]`?

rgieseke:
FWIW, having the error is due to `"strictNullChecks": true,` (or `strict`) - without I don't get the error.

https://www.typescriptlang.org/tsconfig#strictNullChecks

mhkeller:
Great find! So if we remove that and keep the JSDoc types as they are currently, that solves it? Or is it better to change the syntax as you describe above?

rgieseke:
I don't know. The problem occurs because the default of a new SvelteKit project with JSDoc/Typescript includes `strict`: `true`.

Thus it might be better to change the syntax. I'll need to test that, i am slowly wrapping my head around where things go and when to `build` and `package` (hence my initial confusion).

rgieseke:
I think the leading `?` might be in the wrong position (at least for Vscode which uses Typescript's JSDoc extension as far as I understood):

image

There is no difference when they are left out:
image

Maybe they should go here:
image

But maybe we rather want this, only the second parameter being nullable?
image

The domains in the examples all seem to be `[0, null]` (or another array with specific years etc.).

When turning the example above in a little Test module I can import the component without problems and without any `'null' not assignable` errors, but not yet when changing LayerCake's `yDomain` definition similarly.

I found these links helpful:
https://stackoverflow.com/questions/52893679/difference-between-union-types-with-null-and-nullable-object-declaration-in-type
https://www.typescriptlang.org/docs/handbook/2/objects.html#tuple-types

rgieseke:
I switched to the `[min: Number|null, max: Number|null]` variant and tested this locally.

In `partialDomain.js` both values could be null and [null, 0] can be used for data going from negative values to 0.

https://github.com/mhkeller/layercake/blob/main/src/lib/utils/partialDomain.js

mhkeller:
Wow thanks a bunch for diving in and doing all of this great research! I'll merge this and put it as a patch release, which I should be able to do later today.

rgieseke:
Thank _you_! Very glad to be able to give back a tiny bit to LayerCake, it's awesome!

mhkeller:
Are there any changes that should be made to the starter template https://github.com/mhkeller/layercake-template?

mhkeller:
This is published as 7.6.1!

rgieseke:
> Are there any changes that should be made to the starter template https://github.com/mhkeller/layercake-template?

I thought about that, too. I'll take a look.

Issue #151, opened 2023-10-13
Sorted ordinal regression


techniq:
`8.0.0` (PR: https://github.com/mhkeller/layercake/pull/148) appears to have a regression when using `Date` instances - https://github.com/techniq/layerchart/issues/54.

I haven't looked exhaustively, but I think this affects many of the ordinal examples (I could be wrong, but this might be affecting the order of the [Pie](https://www.layerchart.com/docs/components/Pie).  

Is it possible to:
- opt-out of this functionality, or preferable make it opt-in?
- better handle `Date` instances

mhkeller:
Can you make a LayerCake-only repro in the REPL so I can check it out?

techniq:
I just tweaked the column example...  https://svelte.dev/repl/2d4d21d550c446f181b067490fcc25bb?version=3.46.2

Note, if you leave the date's as strings... they are sorted as expected... 

![image](https://github.com/mhkeller/layercake/assets/177476/6bc4cf62-db17-40b8-848e-a12dacc855a8)

but if they are `Date` instances, they sort oddily...

![image](https://github.com/mhkeller/layercake/assets/177476/05846cff-a61f-472a-8092-3fd93756bf69)

So far I've only looked into this Date example... although I'm a little leery of always sorting ordinal data, especially without a way to opt-out (although I don't have a good example at the moment, so maybe it's not an issue in practice).

mhkeller:
Thanks I’ll take a look. You can always override it by setting the xDomain manually.

mattlangeman:
I believe the oddly sorted dates in the example above are getting sorted to the toLocalString representation of the date objects, which starts with day of the week. (Friday, Monday, Saturday, Sunday, Thursday, Tuesday, Wednesday).

https://svelte.dev/repl/b1a6573b77244700af2dd9ac4350abe1?version=4.2.1

mhkeller:
Looks like adding a sort function should fix it https://svelte.dev/repl/28f554b370cf4cc484a5153e257888ca?version=4.2.1

mhkeller:
To say more about whether this is a good default, my thinking was the alternative is it will display in the order it appears in your data, which has no real guarantees and may not be something people think of when trying to debug stuff. I figured a sorted domain was a sensible default but mostly exists to display something halfway decent on first load and people can always override it if they want the domain to be different. 

[...]

PR #152, opened 2023-10-14
Add comparator function to ordinal domain sort

mhkeller:
See https://github.com/mhkeller/layercake/issues/151


mhkeller:
This also adds a fix where unique dates would not be calculated properly. Now, it compares the dates using `.getTime()` instead of relying on `Set` to calculate uniqueness.

src/lib/lib/calcUniques.js
```diff
@@ -35,26 +38,43 @@ export default function calcUniques (data, fields, { sort = false } = {}) {
...
 				for (k = 0; k < vl; k += 1) {
+					/**
+					 * If val[k] is a date, return its time. Otherwise, return val[k].
+					 * Push val[k] to results, though, so that we aren't converting the users data.
+					 */
+					if (!set.has(getTime(val[k]))) {
+						set.add(getTime(val[k]));
+						results.push(val[k]);
+					}
 					set.add(val[k]);
```
mattlangeman, line 62: I don't think this is needed.

src/lib/lib/calcUniques.js
```diff
@@ -35,26 +38,43 @@ export default function calcUniques (data, fields, { sort = false } = {}) {
...
+			} else {
+				/**
+				 * If val is a date, return its time. Otherwise, return val.
+				 * Push val to results, though, so that we aren't converting the users data.
+				 */
+				if (!set.has(getTime(val)) ){
+					set.add(getTime(val));
+					results.push(val);
+				}
 				set.add(val);
```
mattlangeman, line 73: I don't think this is needed.

mattlangeman:
Looks good! I think the two instances of `set.add(val)` that are outside of the if are not necessary, but otherwise this should fix the odd sort behaviour for dates.

src/lib/lib/calcUniques.js
mhkeller: great catch thanks!

src/lib/lib/calcUniques.js
mhkeller: thanks!

mhkeller:
Great thanks for the help in spotting that! I forgot to delete that line from the prior implementation.

src/lib/utils/sorter.js
```diff
@@ -0,0 +1,13 @@
+/**
+ * Sort comparator function
+ * Sort numbers, strings and dates properly in ascending order
+ *
+ * @param {Object} a - First object to compare
+ * @param {Object} b - Second object to compare
+ * @returns {Number} -1, 0, or 1
+ */
+export default function sorter (a, b) {
```
techniq, line 9: Not necessary, but you could also use `d3.ascending` from [d3-array](https://d3js.org/d3-array/sort#ascending) (since we already have a dependency on it.

src/lib/utils/getTime.js
```diff
@@ -0,0 +1,9 @@
+/**
+ * If val is a date, return its time. Otherwise, return val.
+ */
+export default function getTime (val) {
+	if (Object.prototype.toString.call(val) === '[object Date]') {
```
techniq, line 5: Mostly curious, could this be `if (val instanceof Date) {` instead?  Maybe this handles some additional use cases.

techniq (approved):
Thanks @mhkeller so getting to this so quickly.  Looks great

mattlangeman:
My other thought would be to tweak the comments to say something like
```
* Date objects of the same date are not equivalent in a Set
* If val[k] is a date we turn it into a time value with getTime, otherwise getTime returns val[k].
* Push val[k] to results, though, so that we aren't converting the users data.
```

It took me a bit longer than it should have to understand the if block because I was expecting it to be checking if `val[k]` was date.

src/lib/utils/getTime.js
mhkeller: I was reading about the differences and it seems like the `instanceof` approach doesn't work in certain scenarios where the date was created by another frame. It seems like an obscure use case but I look at what underscore does and they use this approach.

techniq: Thanks.  I figure there was some nuance (always is with javascript :) ).

techniq:
FYI, D3 does something with it's [InternMap](https://d3js.org/d3-array/intern#InternMap) / [InternSet](https://d3js.org/d3-array/intern#InternSet) (used for group/rollup/etc).  Could we use those here as well?

src/lib/utils/sorter.js
mhkeller: Yknow, that's a much better solution and what I should have done to begin with.

mhkeller:
@techniq yea that's also a good idea. Another helpful d3-array function that slipped my mind!

mhkeller:
Thanks you both spotting this and your very helpful comments today!

techniq:
Thanks @mhkeller for the quick fix and awesome library!

mattlangeman:
Thanks @mhkeller for the quick fix and thanks @techniq for pointing out the helpful functionality available from d3.

Issue #162, opened 2024-01-23
Update to SvelteKit 2


rgieseke:
As a step towards updating to Svelte 5 (#156) I started looking into updating layercake to use SvelteKit 2.

mhkeller:
Thanks!

mhkeller:
Closed via https://github.com/mhkeller/layercake/pull/163

PR #163, opened 2024-01-23
Update to SvelteKit 2

rgieseke:
This updates Layercake to use SvelteKit 2 (#162).

There were some issues with using `html-minifier` (and `clean-css` in it)  as they use non-browser based functions which are (no longer) supported by the latest Vite used in SvelteKit 2.

There were some workarounds like definining `process.platform` in Vite config but then there were issues with `resolve` (via `uglifiy-js`). I'm not sure whether the html minification is really necessary so I removed it.
Let me know whether it is, also happy to look into alternatives.

This would also update the minimum Svelte requirement as SvelteKit 2 requires Svelte 4 (https://kit.svelte.dev/docs/migrating-to-sveltekit-2)

I have not yet systematically tested everything, just that `dev` and `build` work.


mhkeller:
Thanks for this start! 

Since the library itself has no dependency on Sveltekit, I wonder if the requirement can still svelte 3? It gets into weird territory because the layercake repo itself is a Sveltekit project with its own dependency, but that produces a library that doesn't care. This will all be a moot point with Svelte 5, though. 

I'll take a look over the coming week or so. Just so people know, this is a change purely to how the website is built and doesn't affect what you can use the library with.

rgieseke:
> Since the library itself has no dependency on Sveltekit, I wonder if the requirement can still svelte 3? It gets into weird territory because the layercake repo itself is a Sveltekit project with its own dependency, but that produces a library that doesn't care. This will all be a moot point with Svelte 5, though.

Good point, i hadn't thought about that!

mhkeller:
Thank you for taking the initiative on this @rgieseke !

Issue #171, opened 2024-03-05
Add prop to not sort a calculated domain from ordinal scale


mhkeller:
As of version 8.0.0, when you use an ordinal scale, layer cake calculates the unique values and [sorts them](https://github.com/mhkeller/layercake/issues/147). It could be useful to add a boolean prop like `sortDomainX`, `sortDomainY`, `sortDomainR`, `sortDomainZ` to turn this behavior on or off.

mhkeller:
Closed via https://github.com/mhkeller/layercake/pull/176

PR #176, opened 2024-03-09
Add `x|y|z|rDomainSort` props

mhkeller:
Per https://github.com/mhkeller/layercake/issues/171, this adds `xDomainSort`, `yDomainSort`, `zDomainSort` and `rDomainSort` that will kick in when the scale is ordinal. 

~*This is a technically a breaking change* because it changes the signature of the exported helper function `calcUniques`. . Users that aren't importing that function separately will be fine~

Currently the third argument is an object with the option of `{ sort: true }`. This PR additionally accepts per-field entries:

```js
{  
  sort: Boolean // Overrides field-specific sort options
  x: { sort: Boolean },
  y: { sort: Boolean },
  z: { sort: Boolean },
  r: { sort: Boolean } 
```

The default is unsorted when using the function directly but sorts when used in your chart.

**Tasks**

- [x] Implement in library
- [x] Update tests
- [x] Update docs


mhkeller:
This is ready to be merged. Since it's technically a breaking change and would push it to 9.0.0, I've giving it a moment to see if anything else should go in this release.

PR #182, opened 2024-03-28
Fix comments for Line and Area component

rgieseke:
Maybe there is a better more detailed description possible, but the current one seems to be copy-n-paste from Area-D3 (d3-area is not used in these two here.).

Adds a type information as well to silence a warning.


mhkeller:
Thanks for this! I've updated the REPL examples too.

PR #189, opened 2024-05-02
Update libraries to latest versions and Svelte to next.120

rgieseke:
I tried running the Layercake repo with an update to the Svelte 5 release candidate. I managed to get it to run with `next.120`.

However, I had to disable using the `Nav` component in the main `+layout.svelte`.

Error was
`ReferenceError: Nav is not defined`

I had this error as well when I had tried this with an earlier pre-release (<next.100 IIRC).

The components and examples seem to run fine at first looking around. That's great as it would mean that the library (given Svelte 5's promise for compatibility with the old APIs) should (mostly?) just work.

So I guess it's about getting the website itself to run.

When I update to [next.120](https://github.com/sveltejs/svelte/releases/tag/svelte%405.0.0-next.121) the Layercake website disappears after an initial load.

The console error message is 
```
Uncaught (in promise) TypeError: (intermediate value)(...) is undefined
    Svg Svg.svelte:35
 ```


mhkeller:
Thanks for putting his together. Would be good to keep the Nav in. Does it help if you remove the ts declaration? Weird it’s saying it’s just not defined.

rgieseke:
Sure, and good idea! Removing `lang="ts"` allows the Nav to work.

rgieseke:
next.121 still errors.

mhkeller:
How does it look using just 120?

rgieseke:
120 looked ok and now [123](https://github.com/sveltejs/svelte/releases/tag/svelte%405.0.0-next.123) has fixed the other problem!

I haven't checked every page, but examples, components work. Looking good!

mhkeller:
Very cool thanks! I'll merge this into the svelte-5 branch and then figure out a plan for merging that branch

PR #190, opened 2024-05-03
Make compatible with Svelte 5

mhkeller:
This PR bumps up the versions in the library to make it more easily compatible with Svelte 5. It *does not* update the API to use runes, eliminate the use of stores or get rid of reactive statements.


mhkeller:
`@sveltejs/vite-plugin-svelte@3.1.0` has a peerDep mismatch that will be fixed once 5.0.0 is out. I may wait until 5.0.0 is out anyway.

mhkeller:
Credit to @rgieseke for the help on this one so far

rgieseke:
This will be outdated soon with new Svelte 'next' releases anyways, but I think the peer dependency needs to be `>=5.0.0-next.123` as 121 was breaking Layer Cake.

For more reliable testing I also did a strict versioning `"svelte": "5.0.0-next.123",`. Might be better to add back the caret.

rgieseke:
The `<script lang="ts">` is also in `layercake-template`. Maybe it needs to be changed there as well?

 https://github.com/mhkeller/layercake-template/blob/3c5bd9854b8bee50b9ee61efc9c7cb207545cceb/src/routes/%2Blayout.svelte#L1

https://github.com/mhkeller/layercake/pull/190/commits/39f3081765e8ffa2fba429f947e15e2b03a813df

mhkeller:
Good call. Fixed

mhkeller:
> This will be outdated soon with new Svelte 'next' releases anyways, but I think the peer dependency needs to be `>=5.0.0-next.123` as 121 was breaking Layer Cake.
> 
> For more reliable testing I also did a strict versioning `"svelte": "5.0.0-next.123",`. Might be better to add back the caret.

Added this

mhkeller:
i think this will be a patch version unless there's a good reason why it should be a minor version

mhkeller:
I gotta say, the website now seems a ton faster!

PR #191, opened 2024-05-06
Update prettier and eslint config

rgieseke:
This PR updates the eslint config to those from a current Sveltekit template.

In `.prettierignore` it adds `docs` so that the generated files remain untouched by `npm run format` or `npm run lint`.
This speeds up the formatting process.

Then a run of `npm run format` is included in the PR with the re-formatted files.


mhkeller:
Thanks for the updates. It's tough keeping everything in sync sometimes. 

Of course when it comes to formatting, I'm not sure I'm into all of the changes prettier does. Or maybe it's things that eslint does? 

Maybe we split this up into just eslint and another just for prettier? I'm curious where some of this is coming from. For example, I find it more readable to have these files listed out on separate lines:

Screenshot 2024-05-06 at 17 14 10

I also prefer to not have parentheses around single arguments in arrow functions. Either way, I'd be in favor of updating the libraries to make them work and all that but probably holding off on a mass reformatting since I'd want to review each file individually.

rgieseke:
Sure! I probably should have created an issue first. I'll close this and do that.

PR #194, opened 2024-05-07
Fix #193: clickable docs code samples on Svelte 5

### LINKED ISSUE #193: Cannot click on code files in examples website
Hello! I noticed that ever since this evening I'm not able to click on code examples on the website.

https://github.com/mhkeller/layercake/assets/7550632/09780975-ebea-47ae-b3e8-3540e6f08594

ekzhang:
This fixes #193. It looks like Svelte 5 slightly changes how reactivity works with `$:` on the data prop on a SvelteKit route.

~I added `if (slug !== data.slug)` so that the page content will still update on a navigation to a different frontend page. But this prevents the component from having `active` stuck to be the first code file on the page.~

### DIFF
--- a/src/routes/components/[slug]/+page.svelte
+++ b/src/routes/components/[slug]/+page.svelte
@@ -19,9 +19,9 @@
 	/** @type {import('./$types').PageData} */
 	export let data;
 	let {slug, content, active} = data;
-	$: if (slug !== data.slug) {
-		({slug, content, active} = data);
-	}
+	$: slug = data.slug;
+	$: content = data.content;
+	$: active = data.active;
 
 	function markdownToHtml(text) {
 		return md.render(text);
--- a/src/routes/components/[slug]/+page.svelte
+++ b/src/routes/components/[slug]/+page.svelte
@@ -19,7 +19,9 @@
 	/** @type {import('./$types').PageData} */
 	export let data;
 	let {slug, content, active} = data;
-	$: ({slug, content, active} = data);
+	$: if (slug !== data.slug) {
+		({slug, content, active} = data);
+	}
 
 	function markdownToHtml(text) {
 		return md.render(text);
--- a/src/routes/components/[slug]/+page.svelte
+++ b/src/routes/components/[slug]/+page.svelte
@@ -18,7 +18,6 @@
 
 	/** @type {import('./$types').PageData} */
 	export let data;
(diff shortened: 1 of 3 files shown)


rgieseke:
How about doing this?

```js
export let data;
$: slug = data.slug;
$: content = data.content;
$: active = data.active;
```
It's also three lines but more explicit.

I think the double destructuring might have come after this Sveltekit 2 change:
https://github.com/sveltejs/kit/discussions/5774#discussioncomment-3292707
 
With Svelte 4 the page navigation seems to work even without the first declaration. Not sure if this was necessary for the Layer Cake site (I found this discussion https://twitter.com/geoffrich_/status/1564297325002227713) but this doesn't affect the not working destructuring with Svelte 5.

mhkeller:
Thanks @ekzhang for flagging this so quickly and @rgieseke for having an encyclopedic knowledge of Svelte 2 commits! That Twitter discussion was what prompted that style and I think my logic was to cover my bases.

src/routes/components/[slug]/+page.svelte
```diff
@@ -19,9 +19,9 @@
 	/** @type {import('./$types').PageData} */
 	export let data;
 	let {slug, content, active} = data;
```
Comment (mhkeller) [inline src/routes/components/[slug]/+page.svelte:21]: Is this line still needed?

Comment (mhkeller) [reply src/routes/components/[slug]/+page.svelte:21]: specifically `let {slug, content, active} = data;`

Comment (ekzhang) [reply src/routes/components/[slug]/+page.svelte]: Thanks, it works without the line

mhkeller:
Hm I'm getting this error when I navigate to a component page: 
Screenshot 2024-05-07 at 10 09 17

rgieseke:
> having an encyclopedic knowledge of Svelte 2 commits!

Haha - I did quite a bit of searching to figure out what these commands meant!

Could have been the automatic tool or the recommendation during one of the Sveltekit updates.

In the Twitter thread Rich Harris also wrote:
>[...] the rule is 'only reactive statements can depend on reactive statements', and i think we should avoid promoting the pattern of destructuring `data`. it ends up being more code in 90% of cases, and obscures stuff (if i see `data.foo` in a component i can guess without further context that it comes from a `load` function, etc)

[Twitter](https://twitter.com/Rich_Harris/status/1564344943535181827)

PR #195, opened 2024-05-07
Fix rendering of components pages

ekzhang:
Sorry about not testing the last change thoroughly in #194 — the removal of the `let` line broke the components pages.


mhkeller:
No worries, thanks for the work on this!

Issue #88, opened 2022-07-05
Tweened scales / bounds


techniq:
LayerChart currently has a [Bounds](https://github.com/techniq/layerchart/blob/master/src/lib/components/Bounds.svelte) component that uses a fascade over a d3-scale (see [motionScale()](https://github.com/techniq/layerchart/blob/master/src/lib/utils/scales.ts#L99-L124), which is an abstraction to use either a `tweened()` or `spring()` store.  See also [tweenedScale()](https://github.com/techniq/layerchart/blob/master/src/lib/utils/scales.ts#L45-L70) and [springScale()](https://github.com/techniq/layerchart/blob/master/src/lib/utils/scales.ts#L72-L97) for more direct implementations.

Anyways, `Bounds` is used to "zoom" into a hierarchy chart by adjusting the `domain` and/or `range` of these scales and tweening to the value.   Examples:

- https://www.layerchart.com/examples/Partition (domain only)
- https://www.layerchart.com/examples/Treemap (domain only)
- https://www.layerchart.com/examples/Sunburst (domain and range)

A similar approach can be found on the d3 [zoomable-treemap](https://observablehq.com/@d3/zoomable-treemap), [Pancake](https://pancake-charts.surge.sh/)'s [treemap](https://svelte.dev/repl/2a8acd2328ff4a5398def8b8c8252904?version=3.48.0), and a [many other](https://observablehq.com/search?query=zoomable&onlyOwner=false) d3 zoomable hierarchies.

Pancake's [approach](https://github.com/Rich-Harris/pancake/blob/master/src/lib/components/Chart.svelte#L16-L59) is to allow you to pass the bounds as `x1`, `y1`, `x2`, `y2` and building a new scale reactively (which is basically what `Bounds` does).

Anyways, it would be nice if `<LayerCake>` provided a way to handle this directly, without an extra component / separate scales.  This would also allow to use `$xGet` / `$yGet` / etc instead of a second pair of scales to get the tweened value.

I've been thinking on this for a while, and don't have a specific approach on adding this to LayerCake yet, and feel like it would be fairly deep reaching.  I've considered forking LayerCake instead LayerChart and seeing how it goes.  I mostly created this issue to get your thoughts / feedback, and if you've thought of this.  I remember when I saw LayerCake's [Reactivity Edition](https://github.com/mhkeller/layercake/releases/tag/v6.1.0) I had hopes this might solve this, but it didn't appear to.

[...]

PR #210, opened 2024-06-10
Add `deriveScales` prop

mhkeller:
Adds a prop that lets the user create a new scale based off one or more existing scales.

Usage creating a new scale called `subroupScale` that is the same as the rScale but uses the bandwidth of the xScale as part of its range.

```svelte
deriveScales={{
  subroupScale: ({ xScale, rScale }) => {
    rScale.range([0, xScale.bandwidth()]);
    return rScale;
  }
}}
```


mhkeller:
@techniq if this looks good, i'll merge this in after i do the docs

techniq:
TBH I haven't gotten a chance to kick it around, but the approach looks great and I say ship it.

I'm looking to leverage it at some point within LayerChart, hoping to simplify [createDimensionGetter](https://github.com/techniq/layerchart/blob/main/packages/layerchart/src/lib/utils/rect.ts#L17) and hopefully [improve](https://github.com/techniq/layerchart/issues/98) tooltips for group/stacked data

techniq:
Any chance this will be merged and released in the next day or two? I'm currently overhauling how I handle groups and stacks in LayerChart and would love to leverage this for grouping.

mhkeller:
I wanted to get a couple other eyes on it first to make sure I got the reactivity correct but that hasn't come through yet. Are you able to either `npm link` to this branch or do `npm install mhkeller/layercake#derive-scales` to install from this branch so you can develop locally?

techniq:
> I wanted to get a couple other eyes on it first to make sure I got the reactivity correct but that hasn't come through yet. Are you able to either `npm link` to this branch or do `npm install mhkeller/layercake#derive-scales` to install from this branch so you can develop locally?

I always forget about installing via a git branch.  I just tried via `pnpm add mhkeller/layercake#derive-scales` but ran into issues right out the gate...

image

image

I'm OK with waiting for now as I'd like to release the new simplified charts hopefully in the next 2 weeks, and wouldn't want to have a local-only dependency holding things up (or break the CI PR build).

techniq:
Just had a thought, if you would want to publish the package as a `next` version I could try that it that way.

mhkeller:
@techniq Try installing `8.4.0-beta.1`

techniq:
> @techniq Try installing `8.4.0-beta.1`

Just had a quick minute to install, and looks like I'm in business!

image

I'll start kicking it around this week and let you know how it goes (I'm pretty eager to see how well this cleans up some things, namely [createDimensionGetter](https://github.com/techniq/layerchart/blob/main/packages/layerchart/src/lib/utils/rect.ts#L19C17-L19C38), as moving this up to the context will be much cleaner).

Thanks for the PR and pushing the beta package.

techniq:
Hey @mhkeller, had a little bit of time to play around with this, and ran into some issues / have some suggestions.

- No accessor associated for derived scales (must leverage another scale such as `r` or `z`, as you did in your [example](https://github.com/mhkeller/layercake/blob/99fe091974733b8d550540de0d605ddf4ff45d18/src/routes/_examples/ColumnGrouped.svelte#L52-L67))
- Difficult to know which derived scales MIGHT be defined, which is more of a specific issue for LayerChart but could be solved with conventions (expect `derivedScale={{ x1Scale: ... }}`)
- You can create ANY number of derived scales based on ANY "standard" scale, but in practice I think there is a coupling between a single derived scale and a single standard scale

As with your [example](https://github.com/mhkeller/layercake/blob/99fe091974733b8d550540de0d605ddf4ff45d18/src/routes/_examples/ColumnGrouped.svelte#L52-L67), the only way to associate an accessor with a derived scale is to use/consume another scale such as `r`/`rScale`.  In a lot of ways you are just converting the `rScale` to a derived scale, than creating a new one.

```svelte
<LayerCake
  x="group"
  xScale={scaleBand().paddingInner(0.1).round(true)}
  r="subgroup"
  rScale={scaleBand().paddingInner(0).round(true)}
  deriveScales={{
    subroupScale: ({ xScale, rScale }) => {
      rScale.range([0, xScale.bandwidth()]);
      return rScale;
    }
  }}
>
```

This will make it hard in practice to have more than 1 or 2 derived scales (although I'm not sure how often that will truly be needed), but it also has a tight coupling between 2 scales.

Instead of creating any number of derived scales, I think it might be better to have a single derived scale for each of the current scales (`xScale`, `yScale`, `zScale`, `rScale`), and just pass that scale as a parameter to `x1Domain`, `x1Range`, etc.

```svelte
<LayerCake
  {data}
  x="group"
  xScale={scaleBand().paddingInner(0.1).round(true)}
  x1="subgroup"
  x1Scale={scaleBand().paddingInner(0).round(true)}
  x1Domain={new Set(...data.map(d => d.subgroup))}
  x1Range={xScale => [0, xScale.bandwidth()])}
>
```

This was basically the idea in this [comment](https://github.com/mhkeller/layercake/issues/88#issuecomment-1179823602)), but maybe we don't need to bother with some the extra conventions mentioned (`x1Scale` defaults to `scaleBand()` if `xScale={scaleBand}`.

Thoughts?

techniq:
Hey @mhkeller, I found a way to implement `x1` and `y1` scales in LayerChart by leveraging it's `<ChartContext>`, which I added to improve context types via `chartContext()` which extends `getContext('LayerCake')` but also has better typing, and gives access to LayerCake captured values like `width` and `height`, which is needed for the `x1`/`y1` scales as well as a `resize` event we added [recently](https://github.com/techniq/layerchart/pull/231) (for some force visualization use cases).

You can see the new scales in practice by checking out a few of the [column](https://techniq-simplified-charts.layerchart.pages.dev/docs/examples/Columns#grouped) and [bar](https://techniq-simplified-charts.layerchart.pages.dev/docs/examples/Bars#grouped) examples, but where it really shines, is enabling `seriesLayout="group"` for [BarChart](https://techniq-simplified-charts.layerchart.pages.dev/docs/components/BarChart#group_series).

My plan is to continue to leverage `<ChartContext>` where extensions are needed, for example adding a new `cScale` for colors (instead of (ab)using the `rScale`), and work through the other [Chart improvements](https://github.com/techniq/layerchart/issues/183) ideas.

At some point I would love to consolidate/simplify more logic currently spanned between LayerChart's `<Chart>`/`<ChartContext>` and LayerCake's `<LayerCake>` components where it makes sense (move functionality up/down this stack).

```svelte
<Chart>
  <LayerCake>
    <ChartContext>
      <slot />
    </ChartContext>
  </LayerCake>
</Chart>
```

mhkeller:
Very neat! So I think based on your tests, this approach is generally not the way to go. I'll close this PR as I rethink what the best API is.

mhkeller:
I wonder what the best thing to do about the beta version on npm is. I may end up publishing version 8.0 where extents is deprecated and domain sort order default is changed.

PR #214, opened 2024-06-20
Create new derived store for possibly updated domain

### LINKED ISSUE #213: `$domain` does not reflect `.nice()`
Thanks so much for this great library! I use it nearly daily in my work.

I’m not sure if this is by design, but `$domain` does not reflect the extended domain when applying `.nice()` to the scale. Using `$yScale.domain()` in contrast returns the correct (extended) domain.

Here is an example of this behavior: https://svelte.dev/repl/cfbbdba51fa54921ae8a568b9e529576?version=4.2.18

In this example, the values differ like this:
```
$yDomain: [242.49063670412, 709.433333333333]
$yScale.domain(): [200, 750]
```

jnsprnw:
This is a quick draft and suggestion of how to fix #213.

Resetting `$domain` in [`createScale.js`](https://github.com/mhkeller/layercake/blob/7ccd128ac3184322415f88974d5a184fdf0ba8fd/src/lib/helpers/createScale.js#L42) did not work as the function has no access to the domain store from the upper level.

This solution is not the most elegant, but I couldn’t find another one yet.

### DIFF
--- a/src/_components/Bar.svelte
+++ b/src/_components/Bar.svelte
@@ -9,8 +9,6 @@
 
 	/** @type {String} [fill='#00bbff'] - The shape's fill color. This is technically optional because it comes with a default value but you'll likely want to replace it with your own color. */
 	export let fill = '#00bbff';
-
-	$: console.log({ $xDomain })
 </script>
 
 <g class="bar-group">
--- a/src/_components/Bar.svelte
+++ b/src/_components/Bar.svelte
@@ -5,7 +5,7 @@
 <script>
 	import { getContext } from 'svelte';
 
-	const { data, xGet, yGet, xScale, yScale, xDomain } = getContext('LayerCake');
+	const { data, xGet, yGet, xScale, yScale } = getContext('LayerCake');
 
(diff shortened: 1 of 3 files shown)


mhkeller:
That seems like a pretty light-touch way to do it. Let me think on it a bit if there are any alternatives. Thanks a bunch for digging into it!

mhkeller:
@jnsprnw How's that look now?

techniq:
I ran into this as well and just used `$xScale.domain()`/etc instead :).

- https://github.com/techniq/layerchart/blob/7152aa2d4d94b920fb94bbfcc25fdf9ade732236/packages/layerchart/src/lib/utils/rect.ts#L26-L28
- https://github.com/techniq/layerchart/blob/main/packages/layerchart/src/lib/components/Brush.svelte#L31-L33

Nice to not have to know this and work around.

jnsprnw:
Thanks so much for the quick merge!

mhkeller:
Thanks for the quick PR!

Issue #218, opened 2024-06-26
Typescript errors


rgieseke:
Again following up after #211 (and #191), there are a bunch of Typescript errors when running `npm run check`.

As discussed previously (e.g. #49, #56, #117) it's tricky to get everything typed up properly with Layer Cake providing many generic layers. Still it would help if there were fewer (or no) error messages, especially when including a Layer Cake component in a Typescript based project (so ensure stuff in `src/_components` is error free if possible).

I'm very happy to help with cleaning things up but would be good to get your current thoughts on this, @mhkeller, if you have any.

I already tried experimenting a bit with this, but things get messy easily (e.g. more errors if one starts typing in one place.) I could also try starting with smaller PRs which are probably easier to discuss than general big picture questions.

mhkeller:
From another perspective, components in `src/_components` are starter examples that users can include their project and those would be the least important to add types to whereas the doc site is a done thing. 

There are a bunch of layer cake users who don't use typescript and so I would be worried that it would make these starter components less approachable for them if we included a bunch of type annotations that they're not used to seeing. At the same time, users who want those annotations aren't blocked on adding types to satisfy their linters etc. 

If there's a middle ground I'm missing, though, feel free to start a PR with what you think is the best way to handle. Overall, wrestling with typescript isn't that appealing to me personally but if there are improvements that are easy to do then let's take a look.

mhkeller:
I appreciate the help and enthusiasm – also helpful for me to learn things I otherwise wouldn't!

mhkeller:
Dropping this cheatsheet in here for reference: https://alexharri.com/blog/jsdoc-as-an-alternative-typescript-syntax

@rgieseke I thought the PR for the axes was a good first step. Are you interested in adding any more? I can slowly make my way through, as well.

rgieseke:
> @rgieseke I thought the PR for the axes was a good first step. Are you interested in adding any more? I can slowly make my way through, as well.

I definitely am interested in adding more! I'll try to keep you posted when I start on something so we don't have duplicated efforts.

mhkeller:
Great! What may be easiest to avoid duplicate work is to issue PRs on smaller batches of files and merge them into a branch called `ts-fixes` or something instead of into `main` to avoid having to do multiple releases and `main` diverging a lot from the published npm version in the mean time. Open to any other ideas, though.

mhkeller:
[...]

PR #221, opened 2024-06-26
Fix TypeScript errors in Axis components

rgieseke:
Here are fixes for the Axis components. I think they wouldn't be too overwhelming for non-Typescript users as discussed in #218  

If you're fine with that level of additional typing I could try and see where else I could make similar changes (probably in bits whenever I have some time).

One thing I'm not sure about in this PR is the whether this is correct. Can the Array only contain numbers, or also potentially strings?

```js
/** @type {Number|Array<Number>|Function|undefined} [ticks] - If this is a number, it passes that along to the [d3Scale.ticks](https://github.com/d3/d3-scale) function. If this is an array, hardcodes the ticks to those values. If it's a function, passes along the default tick values and expects an array of tick values in return. If nothing, it uses the default ticks supplied by the D3 function. */
export let ticks = undefined;
```

### DIFF
--- a/src/_components/AxisX.percent-range.html.svelte
+++ b/src/_components/AxisX.percent-range.html.svelte
@@ -27,7 +27,7 @@
 	/** @type {(d: any) => string} [format=d => d] - A function that passes the current tick value and expects a nicely formatted value in return. */
 	export let format = d => d;
 
-	/** @type {Number|Array<Number>|Function|undefined} [ticks] - If this is a number, it passes that along to the [d3Scale.ticks](https://github.com/d3/d3-scale) function. If this is an array, hardcodes the ticks to those values. If it's a function, passes along the default tick values and expects an array of tick values in return. If nothing, it uses the default ticks supplied by the D3 function. */
+	/** @type {Number|Array<any>|Function|undefined} [ticks] - If this is a number, it passes that along to the [d3Scale.ticks](https://github.com/d3/d3-scale) function. If this is an array, hardcodes the ticks to those values. If it's a function, passes along the default tick values and expects an array of tick values in return. If nothing, it uses the default ticks supplied by the D3 function. */
 	export let ticks = undefined;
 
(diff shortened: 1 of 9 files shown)


src/_components/AxisX.percent-range.html.svelte
```diff
@@ -24,10 +24,10 @@
 	/** @type {Boolean} [snapLabels=false] - Instead of centering the text labels on the first and the last items, align them to the edges of the chart. */
 	export let snapLabels = false;
 
-	/** @type {Function} [format=d => d] - A function that passes the current tick value and expects a nicely formatted value in return. */
+	/** @type {(d: any) => string} [format=d => d] - A function that passes the current tick value and expects a nicely formatted value in return. */
```
rgieseke, line 27: Maybe `any` could be more precisely defined.

mhkeller:
These look great. I had thought we'd have to do more inline type definitions such as `function thing (d /* type stuff here*/) { }` which would visually throw some folks off.

rgieseke:
> These look great. I had thought we'd have to do more inline type definitions such as function thing (d /* type stuff here*/) { } which would visually throw some folks off.

Yes, fully agreeing!

I need to give this more thought, I think the different `ticks`, `tickVals` and `format` function types need to be properly aligned.

src/_components/AxisX.percent-range.html.svelte
rgieseke: It's probably o.k. to keep these values as `any` then, I've also changed the other places of `Array<Number>` to `Array<any>` since we can't really know what users might use in their scale functions.

src/_components/AxisX.svelte
```diff
@@ -83,6 +86,7 @@
 					y2={tickGutter + tickLen}
 				/>
 			{/if}
+			@param {Number} i/** @param {Boolean} sl */
```
mhkeller, line 89: What does this one refer to?

rgieseke: Whoops! Accidental paste I guess.

Issue #217, opened 2024-06-26
svelte-check error from svelte-app directory


rgieseke:
When running `npm run check` I'm getting

```
> layercake@8.3.3 check
> svelte-kit sync && svelte-check --tsconfig ./jsconfig.json


====================================
Loading svelte-check in workspace: /home/robert/Code/layercake
Getting Svelte diagnostics...

Error while loading config at  /home/robert/Code/layercake/src/scripts/svelte-app/svelte.config.js
Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@sveltejs/adapter-auto' imported from /home/robert/Code/layercake/src/scripts/svelte-app/svelte.config.js
    at packageResolve (node:internal/modules/esm/resolve:844:9)
    at moduleResolve (node:internal/modules/esm/resolve:901:20)
    at defaultResolve (node:internal/modules/esm/resolve:1131:11)
    at ModuleLoader.defaultResolve (node:internal/modules/esm/loader:390:12)
    at ModuleLoader.resolve (node:internal/modules/esm/loader:359:25)
    at ModuleLoader.getModuleJob (node:internal/modules/esm/loader:234:38)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/module_job:85:39)
    at link (node:internal/modules/esm/module_job:84:36) {
  code: 'ERR_MODULE_NOT_FOUND'
}
```

I couldn't figure out why `svelte-check` is trying to load the additional config file as the directory should be ignored from the `exlude` pattern in jsconfig.json.

https://github.com/mhkeller/layercake/blob/65011709df138a86085cab744b70c339402751bb/jsconfig.json#L26-L27

`adapter-auto` is not installed and installing `@sveltejs/adapter-auto` removes the error but I'm not sure that is the 'correct' solution.

As i understand the `update_template.sh` script is used to create the download button content, right?

mhkeller:
Seems like Typescript `exclude` doesn't actually exclude things: https://github.com/Microsoft/TypeScript/issues/24133

mhkeller:
You could do some workaround where you add a `// @ts-nocheck` when the template repo gets cloned and then remove it when the user downloads it but that seems like a lot of moving parts to satisfy the build tool. im kinda inclined to ignore this since.

mhkeller:
Thanks for taking it on. Looks like a good workaround.

PR #222, opened 2024-06-27
Workaround for svelte-check error from additional config file

rgieseke:
The app template in `src/scripts/svelte-app` has another `svelte.config.js` which causes an error message when running `npm run check` as `adapter-node` is not installed. To workaround the file is renamed and then added separately with the correct name in `static/svelte-app.json`.

See #217


mhkeller:
Thanks for putting this together.

PR #223, opened 2024-06-28
Fix TypeScript errors from missing undefined or unprecise arrays

rgieseke:
Some time on a train, so here comes another TypeScript error message fix, basically I added `|undefined` to the type information whenever `undefined` is actually set.

This also how it's done in Svelte components itself, so I'm quite confident that that is reasonable.

```svelte
/** @type {import('./types').MessageDetails | undefined} */
export let details = undefined;

/** @type {string | undefined} */
export let filename = undefined;
```	

https://github.com/sveltejs/svelte/blob/33e44ea697356ea3f21fc6649919cf7fb4948b22/sites/svelte-5-preview/src/lib/Message.svelte#L8

I also turned some arrays into arrays of objects.


mhkeller:
Great. I made some adjustments to the docs so the changes are reflected there, too. For `data` and `flatData`, I made those `Array<any>` since you could have arrays of arrays like in a stacked data structure.

mhkeller:
Let me know if this is done. I'll put it through as a patch release since it improves the types on the LayerCake component itself.

rgieseke:
Wicked, thank you. Looks good to me!
Is there a script to update the docs (or way to check they match)?

mhkeller:
Nope I don't have anything automated. Also what's not automated are the components in the REPL since I don't think there's any way to programmatically access svelte REPLs.

PR #226, opened 2024-08-15
Lower-case type synonym for Number

rgieseke:
See discussion in #218, 

From https://github.com/mhkeller/layercake/issues/218#issuecomment-2287192896

> For legacy reasons, Number in JSDoc is interpreted as number, along with string and boolean.

To avoid confusion (especially for users coming from TypeScript) the lower-case variants are used.

### DIFF
--- a/src/_components/AxisRadial.svelte
+++ b/src/_components/AxisRadial.svelte
@@ -7,10 +7,10 @@
 
 	const { width, height, xScale, extents, config } = getContext('LayerCake');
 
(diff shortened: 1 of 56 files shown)


rgieseke:
This is already quite a bit to scroll through for review so I'll do other type synonyms in separate PRs for easier review.

mhkeller:
Sorry I missed this notification. Thanks for going through it!

PR #229, opened 2024-08-26
Add type information for mousemove event

rgieseke:
This adds type information for the mousemove event in the Quadtree components. The error was:

```
src/_components/QuadTree.html.svelte:32:20
Error: Parameter 'evt' implicitly has an 'any' type. (js)

	function findItem(evt) {
		e = evt;
--
src/_components/QuadTree.percent-range.html.svelte:32:20
Error: Parameter 'evt' implicitly has an 'any' type. (js)

	function findItem(evt) {
		e = evt;
```

This also adds a JSDoc typecast for the 'layerX'/'layerY' values which probably is an example of those type annotations you didn't like so much (and me neither).
I didn't find another way to let the checking accept the uppercase conversion in `layer${x.toUpperCase()` as a type of `'layerX'|'layerY'`.

### DIFF
--- a/src/_components/QuadTree.html.svelte
+++ b/src/_components/QuadTree.html.svelte
@@ -14,10 +14,10 @@
 	let found = {};
 	let e = {};
 
-	/** @type {String} [x='x'] – The dimension to search across when moving the mouse left and right. */
+	/** @type {'x'|'y'} [x='x'] – The dimension to search across when moving the mouse left and right. */
 	export let x = 'x';
 
-	/** @type {String} [y='y'] – The dimension to search across when moving the mouse up and down. */
+	/** @type {'x'|'y'} [y='y'] – The dimension to search across when moving the mouse up and down. */
 	export let y = 'y';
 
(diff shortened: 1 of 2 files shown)


src/_components/QuadTree.html.svelte
```diff
@@ -29,12 +29,13 @@
...
+	/** @param {MouseEvent} evt */
 	function findItem(evt) {
 		e = evt;
 
-		const xLayerKey = `layer${x.toUpperCase()}`;
-		const yLayerKey = `layer${y.toUpperCase()}`;
-
+		const xLayerKey = /** @type {'layerX'|'layerY'} */ (`layer${x.toUpperCase()}`);
+		const yLayerKey = /** @type {'layerX'|'layerY'}*/ (`layer${y.toUpperCase()}`);
+		console.log(evt.x, evt.y);
```
mhkeller, line 38: I'm going to remove this log

mhkeller:
That makes sense. Ya I tend to find those inline types harder for newcomers to read but seems like this is the best solution.

src/_components/QuadTree.html.svelte
rgieseke: Ooop, sory! Thanks for the merge!

PR #230, opened 2024-08-26
[Draft] Ts fixes

mhkeller:
Per https://github.com/mhkeller/layercake/issues/218

Collecting Typescript fixes on this branch


mhkeller:
@rgieseke should we merge this in? there are definitely more typescript changes that could be made but id like to start condensing all of the work we've done and it's kind of pointless to fix a bunch of ts errors before github.com/mhkeller/layercake/pull/271 and https://github.com/mhkeller/layercake/pull/281

I'm looking into why the CodeQL is currently failing.

rgieseke:
Yeah, we can always improve types further later!

PR #231, opened 2024-08-29
Fix TypeScript errors in lib/utils

rgieseke:
This fixes a few more TypeScript errors, those in `lib/utils` (see #218).

The error messages were:

```
src/lib/utils/makeAccessor.js:4:33
Error: Generic type 'Array<T>' requires 1 type argument(s). 
	Make an accessor from a string, number, function or an array of the combination of any
	@param {string|number|Function|Array} acc The accessor function, key or list of them.
	@returns {Function} An accessor function.
--
src/lib/utils/makeAccessor.js:8:23
Error: Type 'null' is not assignable to type 'Function'. 
export default function makeAccessor(acc) {
	if (!canBeZero(acc)) return null;
	if (Array.isArray(acc)) {
--
src/lib/utils/makeAccessor.js:10:10
Error: Parameter 'd' implicitly has an 'any' type. 
	if (Array.isArray(acc)) {
		return d =>
			acc.map(k => {
--
src/lib/utils/makeAccessor.js:15:10
Error: Parameter 'd' implicitly has an 'any' type. 
	} else if (typeof acc !== 'function') {
		return d => d[acc];
	}
--
src/lib/utils/filterObject.js:10:34
Error: Element implicitly has an 'any' type because expression of type 'string' can't be used to index type '{}'.
  No index signature with a parameter of type 'string' was found on type '{}'. 
		Object.entries(obj).filter(([key, value]) => {
			return value !== undefined && comparisonObj[key] === undefined;
--
src/lib/utils/debounce.js:8:6
Error: Variable 'timer' implicitly has type 'any' in some locations where its type cannot be determined. 
export default function debounce(func, timeout = 300) {
	let timer;
	return (...args) => {
--
src/lib/utils/debounce.js:9:10
Error: Rest parameter 'args' implicitly has an 'any[]' type. 
	let timer;
	return (...args) => {
		clearTimeout(timer);
--
src/lib/utils/debounce.js:10:16
Error: Variable 'timer' implicitly has an 'any' type. 
	return (...args) => {
		clearTimeout(timer);
		timer = setTimeout(() => {
--
src/lib/utils/debounce.js:12:15
Error: 'this' implicitly has type 'any' because it does not have a type annotation. 
		timer = setTimeout(() => {
			func.apply(this, args);
		}, timeout);
--
src/lib/utils/arraysEqual.js:6:10
Error: Generic type 'Array<T>' requires 1 type argument(s). 
	of making a set
	@param {Array} arr1 An array to test
	@param {Array} arr2 An array to test against
--
src/lib/utils/arraysEqual.js:7:10
Error: Generic type 'Array<T>' requires 1 type argument(s). 
	@param {Array} arr1 An array to test
	@param {Array} arr2 An array to test against
	@returns {boolean} Whether they contain all and only the same items
--
src/lib/utils/padScale.js:19:34
Error: Parameter 'scale' implicitly has an 'any' type. 

export default function padScale(scale, padding) {
	if (typeof scale.range !== 'function') {
--
src/lib/utils/padScale.js:19:41
Error: Parameter 'padding' implicitly has an 'any' type. 

export default function padScale(scale, padding) {
	if (typeof scale.range !== 'function') {
--
src/lib/utils/padScale.js:41:38
Error: Parameter 'd' implicitly has an 'any' type. 

	const [d1, d2] = scale.domain().map(d => {
		return isTime ? lift(d.getTime()) : lift(d);

```


src/lib/utils/padScale.js
```diff
@@ -1,21 +1,37 @@
+import isOrdinalDomain from '../helpers/isOrdinalDomain.js';
+import getPadFunctions from '../helpers/getPadFunctions.js';
+import findScaleName from '../helpers/findScaleName.js';
+
+// These scales have a discrete range so they can't be padded
+const unpaddable = ['scaleThreshold', 'scaleQuantile', 'scaleQuantize', 'scaleSequentialQuantile'];
+
+/**
+ * @typedef {import('d3-scale').ScaleLinear<any, any> |
```
rgieseke, line 9: Here I went through the d3-docs and added the interface definitions.

Not sure whether one should exclude those that are filtered out with `unpaddable` or whether there is already combinations like `scaleContinuousNumeric` which would make the list shorter.

src/lib/utils/debounce.js
```diff
@@ -1,15 +1,18 @@
 /**
 	A simple debounce function taken from here https://www.freecodecamp.org/news/javascript-debounce-example/
-	@param {function} func The function to debounce.
+	@param {Function} func The function to debounce.
 	@param {number} timeout The time in ms to wait.
-	@returns {function}
+	@returns {Function}
 */
-export default function debounce(func, timeout = 300) {
+const debounce = (func, timeout = 300) => {
```
rgieseke, line 7: Changing this to an arrow function was the solution I found to `this` not being defined.

mhkeller:
Thanks for this one. I was away for a few weeks and will go through this.

src/lib/utils/debounce.js
mhkeller: Hm, in an arrow function, the reference of `this` changes. I imagine it still works because `this` isn't actually relevant for this call?

rgieseke: Hm, I'm not sure I fully understand the "An outer value of 'this' is shadowed by this container." error message that was there. 

Another approach might be to use something like here: https://decipher.dev/30-seconds-of-typescript/docs/debounce/

With JSDoc:

```javascript
export default function debounce(func, timeout = 300) {
	/** @type {string | number | NodeJS.Timeout | undefined} */
	let timer;
	return function (/** @type {any} */ this, /** @type {any[]} */...args) {
		clearTimeout(timer);
		timer = setTimeout(() => {
			func.apply(this, args);
		}, timeout);
	};
}
```

mhkeller: Got it – I imagine that error is saying that because `this` exists in the parent scope, any child `this` is shadowed by it. I have a feeling it's not a big deal. I'll take a look at this other function, too, for good measure.

Sorry this has taken me longer than usual to review. I am on it, though, and really appreciate the work you've put into it!

rgieseke: Absolutely no worries about reviews taking long! None of this is urgent and I appreciate your careful reviews.

In any case, might also be an option to silence that warning message. I think the main benefit of having a cleanly linted version is to easily see that refactoring or new features don't introduce new problems.

mhkeller: Totally agree on that point about the value it brings.

src/lib/utils/debounce.js
```diff
@@ -1,15 +1,18 @@
...
+	/** @type {string | number | NodeJS.Timeout | undefined} */
```
rgieseke, line 8: This might be more readable (via [Stackoverflow](https://stackoverflow.com/questions/51040703/what-return-type-should-be-used-for-settimeout-in-typescript):

```suggestion
	/** @type {ReturnType<typeof setTimeout>} */
```

src/lib/utils/padScale.js
```diff
@@ -41,7 +57,7 @@ export default function padScale(scale, padding) {
 	const [d1, d2] = scale.domain().map(d => {
 		return isTime ? lift(d.getTime()) : lift(d);
 	});
-	const [r1, r2] = scale.range();
+	const [r1, r2] = /** @type {any[]} */ (scale.range());
```
mhkeller, line 60: Should this be `number[]` instead of `any[]`?

src/lib/utils/padScale.js
mhkeller: I think, even though the script will handle an unpaddable scale, that it should be a typescript error if you try to pass in an unpaddable scale. 

I didn't follow the second part of what you were saying about `scaleContinuous Numeric`. Can you explain what you mean more?

mhkeller:
I think this is looking really good. I made some changes and added some comments. I still need to go through `makeAccessor`. It may be an improvement to define the `Function` more.

src/lib/utils/padScale.js
rgieseke: Sorry I should have linked that. in d3-typed there is [scaleContinousNumeric](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/5d86c5afed68b6dfdaa07fcec9bf9b99325410a2/types/d3-scale/index.d.ts#L32), which then gets extended to `ScaleLinear`, `ScalePower`, `ScaleLogarithmic` and a few more.

src/lib/utils/padScale.js
rgieseke: Yes, `[number, number]` even.

src/lib/utils/padScale.js
```diff
@@ -1,21 +1,37 @@
...
 	the min/max by taking the desired difference
 	in pixels and converting it to units of data.
 	Returns an array that you can set as the new domain.
 	Padding contributed by @veltman.
 	See here for discussion of transforms: https://github.com/d3/d3-scale/issues/150
-	@param {Function} scale A D3 scale funcion
-	@param {number[]} padding A two-value array of numbers specifying padding in pixels
+	@param {Scale} scale A D3 scale funcion
+	@param {[number, number]} padding A two-value array of numbers specifying padding in pixels
 	@returns {number[]} The padded domain
```
mhkeller, line 30: The return type is `[number, number]` also, right?

rgieseke:
I hadn't had a chance to work on Layer Cake projects for a while, but now I was working on one and am motivated to finish this!

Happy to simplify this into multiple PRs, if there is some stuff that still needs more thinking and sorry for the long delay!

mhkeller:
I think keeping it in one PR is okay. I think I needed to still go through some of the files, too. I've been spending a lot of time on a layercake-adjacent project but the positive part of that is I have been using types a lot more. I appreciate your help in working on this more!

PR #237, opened 2024-10-04
Add types to range calculation functions

rgieseke:
Maybe the `s` parameter should only be `y` or `r`?

### DIFF
--- a/src/lib/settings/getDefaultRange.js
+++ b/src/lib/settings/getDefaultRange.js
@@ -1,3 +1,10 @@
+/**
+ * @param {'x'|'y'|'z'|'r'} s
+ * @param {number} width
+ * @param {number} height
+ * @param {boolean} reverse
+ * @param {boolean} percentRange
+ */
 function calcBaseRange(s, width, height, reverse, percentRange) {
 	let min;
 	let max;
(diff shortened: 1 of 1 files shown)


mhkeller:
What's your thinking in having it just be those two? `createScale` gets called with all of those parameters, no?

src/lib/settings/getDefaultRange.js
```diff
@@ -11,6 +18,14 @@ function calcBaseRange(s, width, height, reverse, percentRange) {
 	return reverse === true ? [max, min] : [min, max];
 }
 
+/**
+ * @param {'x'|'y'|'z'|'r'} s
+ * @param {number} width
+ * @param {number} height
+ * @param {boolean} reverse
+ * @param {[number, number]|Function|null} range
```
mhkeller, line 26: Instead of function, what if we were more specific and defined the arguments. Something like this?

```js
/*
 * @typedef {function} RangeType
 * @param {{width: number, height: number}} dimensions - The dimensions object containing width and height.
 * @returns {Array<number|string>} The calculated range .
```

src/lib/settings/getDefaultRange.js
```diff
@@ -11,6 +18,14 @@ function calcBaseRange(s, width, height, reverse, percentRange) {
...
+ * @param {boolean} percentRange
```
mhkeller, line 27: We should probably add a return description to the type like in the above comment

rgieseke:
> What's your thinking in having it just be those two? createScale gets called with all of those parameters, no?

Uh, yeah, I got confused in the ternaries handling `min`/`max` - 'x' and 'z' are the final case.

mhkeller:
Handled in https://github.com/mhkeller/layercake/pull/231

Issue #147, opened 2023-09-08
Should auto-calculated ordinal domain return sorted?


mhkeller:
Currently, if you set an accessor for x y or z and you pass in an ordinal scale, the domain will be [calculated as unique elements](https://github.com/mhkeller/layercake/blob/76b2d37aa151567e8573c40e39d9b590c75cfd81/src/lib/helpers/calcScaleExtents.js#L23), say, for a color scale. The order of these unique elements is just based on when they appear in the data. Should `.sort()` be called on them so that the appearance is more consistent?

This could be a potential breaking change for some charts.

mhkeller:
Yea I think this would be a good idea.

mhkeller:
Implemented in 8.0.0

mhkeller:
I'm re-opening this since some people have found this behavior surprising. Maybe better to turn this off by default. Any thoughts welcome

mhkeller:
@techniq @rgieseke @jtrim-ons any preference on this one?

techniq:
I personally find it [surprising](https://github.com/mhkeller/layercake/issues/151), and while `yDomain={[...new Set(data.map((d) => d.name))]}` does work, I think reverting to the `pre-8.0.0` state would be a better default IMO, but not major if you don't.

rgieseke:
I have no strong preference here, thanks for the heads-up!
Not sure if there is an example in the gallery which demonstrates this already?

mhkeller:
@techniq For what it's worth, since [8.1.0](https://github.com/mhkeller/layercake/blob/main/CHANGELOG.md#810) you should be able to set `xDomainSort={false}` and avoid looping through your data and getting the uniques.

PR #241, opened 2024-10-19
Change default sort option to `false` for ordinal scales

mhkeller:
Per https://github.com/mhkeller/layercake/issues/147


mhkeller:
@techniq Could you give this a try and see if that solves your use case?

techniq:
@mhkeller as in, I can remove [this](https://github.com/techniq/layerchart/blob/main/packages/layerchart/src/lib/components/Chart.svelte#L304-L318) that I added in the [0.50.0](https://github.com/techniq/layerchart/blob/eed47b916aba3f319598e2289115849316bc3819/packages/layerchart/CHANGELOG.md?plain=1#L141) release 😁

mhkeller:
Haha yes finallly. Seems like a straightforward change!

PR #258, opened 2025-03-17
Remove half-working ARIA code and A11Y warning

rgieseke:
While looking at converting components to Svelte 5 syntax I saw some warnings in the current versions.
 
For the Bruch component, running `svelte-check` reports a missing `aria-valuenow` attribute for the range slider which only has lower and upper value and no actual 'value', so maybe the role 'slider' is not applicable.

```js
/src/_components/Brush.html.svelte:99:2
Warn: Elements with the ARIA role "slider" must have the following attributes defined: "aria-valuenow" (svelte)
	on:touchstart|stopPropagation={reset}
	role="slider"
	aria-valuemin={min}


./src/_components/Brush.html.svelte:111:4
Warn: Elements with the ARIA role "slider" must have the following attributes defined: "aria-valuenow" (svelte)
			style="left: {left}%; right: {right}%"
			role="slider"
			aria-valuemin={min}


./src/_components/Brush.html.svelte:122:4
Warn: Elements with the ARIA role "slider" must have the following attributes defined: "aria-valuenow" (svelte)
			style="left: {left}%"
			role="slider"
			aria-valuemin={min}


./src/_components/Brush.html.svelte:133:4
Warn: Elements with the ARIA role "slider" must have the following attributes defined: "aria-valuenow" (svelte)
			style="right: {right}%"
			role="slider"
			aria-valuemin={min}
```

The current code wasn't actually keyboard accessible, so one could tab to the component but not change the values with the keyboard.

image

I'd suggest to remove the code for now and silence the warning. It's probably better to have no keyboard accessibility than a half-working one.

### DIFF
--- a/src/_components/Brush.html.svelte
+++ b/src/_components/Brush.html.svelte
@@ -91,6 +91,7 @@
 	$: right = 100 * (1 - max);
 </script>
 
+<!-- TODO Add keyboard accessibility. See https://github.com/mhkeller/layercake/pull/258 -->
 <!-- svelte-ignore a11y_no_static_element_interactions -->
 <div
 	bind:this={brush}
--- a/src/_components/Brush.html.svelte
+++ b/src/_components/Brush.html.svelte
@@ -91,16 +91,12 @@
 	$: right = 100 * (1 - max);
 </script>
 
+<!-- svelte-ignore a11y_no_static_element_interactions -->
 <div
 	bind:this={brush}
 	class="brush-outer"
 	on:mousedown|stopPropagation={reset}
 	on:touchstart|stopPropagation={reset}
-	role="slider"
-	aria-valuemin={min}
-	aria-valuemax={max}
-	aria-valuetext="{min} to {max}"
-	tabindex="0"
 >
 	{#if min !== null}
 		<div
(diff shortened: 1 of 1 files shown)


mhkeller:
That makes sense. Let's add a link to this PR so people can understand what's going on in case they want to add better accessibility to it.

src/_components/Brush.html.svelte
```diff
@@ -91,50 +91,31 @@
 	$: right = 100 * (1 - max);
 </script>
```
mhkeller, line 93: Let's add a comment here

rgieseke:
Good point! I've added the following above the `svelte-ignore`

`<!-- TODO Add keyboard accessibility. See https://github.com/mhkeller/layercake/pull/258 -->`

I looked into accessible brush components but didn't find anything right away . I think it's possible to improve this and add keyboard accessibility.

tannerjaime:
I'm surprised that svelte-check wouldn't accept `aria-valuetext` in place of `aria-valuenow`!

FWIW I'm not sure that I mind having the `aria-valuetext` currently, as it does give screen reader users a bit of info as to whats on the page, but agree that it's not usable by keyboard and assistive tech as is right now so I also wouldn't mind removing and rethinking it. I'm not sure that the slider container needs `role="slider"` for example, and if only the handles on the brush had `role="slider"`, the value of `aria-valuenow` for each would be more intuitive. 

It's not a perfect example, but I think [this input range example](https://www.w3.org/WAI/ARIA/apg/patterns/slider-multithumb/examples/slider-multithumb/) could be a good reference when making improvements to the pattern here in the future.

mhkeller:
Thanks everyone! This is a helpful improvement and if there is a way to make this component keyboard accessible in the future that would be great!

PR #268, opened 2025-04-14
Add aria-label to GitHub icon link

rgieseke:
Warning was:
> src/routes/_site-components/Nav.svelte:118:3 Buttons and links should either contain text or have an `aria-label` or `aria-labelledby` attribute https://svelte.dev/e/a11y_consider_explicit_label

I wasn't sure whether it should describe the logo or the link target, but it seems to be the target:

https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA8


mhkeller:
Thanks!

PR #270, opened 2025-04-22
Add test for code sample navigation and download

rgieseke:
This depends on #261 

It adds a test for navigating between the code files and should catch problems like #193 

It also adds a test to check that the download works and the zip file is not empty. It checks that the file size is not zero and in the range of the expected file size. Since there is no unzip library included this seemed a good-enough smoke test to me.
An alternative I considered was to check for text in the output file (as it's not actually compressed by do-not-zip).


.github/workflows/node.js.yml
```diff
@@ -26,4 +26,4 @@ jobs:
           node-version: ${{ matrix.node-version }}
           cache: 'npm'
       - run: npm ci --legacy-peer-deps
-      - run: npm test
+      - run: npm run test:lib
```
mhkeller, line 29: Now that that there is `test:lib` and `test:playwright`, do we not want both to run in the CI?

rgieseke: 'test:lib' runs for multiple Node versions. For 'test:playwright' we can know which Node version is running in CI and the tested output ist static html so I think the current setup makes sense.

We should think about testing other browsers in the future but probably get one running first (the screenshot test will get harder to match with Firefox).

mhkeller: Ah I had missed that playwright.yml runs as a separate action on CI. I had just given a quick look at this diff

test/e2e/code-examples.spec.js
```diff
@@ -0,0 +1,36 @@
...
+import { existsSync, statSync } from 'fs';
+
+const groups = ['example', 'example-ssr'];
+
+groups.forEach(path => {
+	test(`Navigate between code samples for "${path}/Bar"`, async ({ page }) => {
+		await page.goto(`${path}/Bar`);
+		// Using has-text here to avoid the ul intercepting the click
+		// https://github.com/microsoft/playwright/issues/12821#issuecomment-1069630780
+		await page.click(`li:has-text('./_components/Bar.svelte')`);
```
mhkeller, line 11: FYI I had to change the way the click was being created here because the test was failing the `ul` element was intercepting the click for some reason. This solution seemed maybe better than doing `{force: true}`.

rgieseke: Good to know, thanks!

Issue #156, opened 2023-12-26
Migrate to Runes


saabi:
After attempting to port a working Svelte 3/LayerCake app (using `npm i -S --force LayerCake`) I got the error above (`ERR_SVELTE_TOO_MANY_UPDATES`).

To verify the bug wasn't on my side, I moved an arbitrary working LayerCake REPL example to the Svelte 5 Preview REPL and I get the same error:

```
playground:output:801 Uncaught Error: ERR_SVELTE_TOO_MANY_UPDATES: Maximum update depth exceeded. This can happen when a reactive block or effect repeatedly sets a new value. Svelte limits the number of nested updates to prevent infinite loops.
    at infinite_loop_guard (playground:output:801:10)
    at flush_queued_effects (playground:output:818:4)
    at process_microtask (playground:output:845:3)
```

Working REPL Example: https://svelte.dev/repl/ad32f9f301484d0cace82272126e9d09?version=3.46.2
[...]

Issue #250, opened 2025-01-24
Canvas & Svelte 5 infinite effect loops


cholwell:
All the examples in the docs use $:, if you want to use effect and also do something like `$ctx.strokeStyle` you get an infinite effect loop 

If there is a way to do this properly it would be good to add it to the docs

mhkeller:
Although the library itself is built with Svelte 5, it hasn't yet been updated to use runes. See here for more info: https://github.com/mhkeller/layercake/issues/156. The short version is this library is a collection of stores and will likely need a full rewrite using runes. The reason why that hasn't been done is a) if I migrate to runes, Svelte 4 loses support b) I haven't had the bandwidth. I was aiming for migrating it closer to when Svelte 6 came out and I'm guessing support for Svelte 3 and 4 features is dropped.

In terms of solving your immediate problem, there's a [note in the docs about a related issue with infinite loops](https://layercake.graphics/guide#canvas) which may be relevant. See also https://github.com/mhkeller/layercake/issues/50

@techniq has done some cool stuff overhauling canvas support here https://github.com/techniq/layerchart/pull/295 but I haven't yet had the chance to see how that approach would be relevant to layercake.

techniq:
The solution for LayerChart was to [implement](https://github.com/techniq/layerchart/pull/295) a component registration system where each primitive component [registers](https://github.com/techniq/layerchart/blob/main/packages/layerchart/src/lib/components/Circle.svelte#L86-L97) a `render` method and `Canvas` is responsible for (re)rendering as needed (when `invalidate()` is triggered).  This was released in [0.70.0](https://github.com/techniq/layerchart/releases/tag/layerchart%400.70.0) and improved over subsequent releases.  See also this Bluesky [post](https://bsky.app/profile/techniq.dev/post/3lfn3b4iuec2c) for a breakdown.  There was a lot of reactivity trickiness to nail down (such as [creating](https://github.com/techniq/layerchart/blob/228537f1c7bf9015e55254a6c2a8e31dde2c485b/packages/layerchart/src/lib/components/Circle.svelte#L68-L70) an object Id for [linearGradient](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/createLinearGradient) objects) and managing context save/restore to enable `Group` transforming.  I still need to [solve](https://github.com/techniq/layerchart/issues/307) `ClipPath` and a few other small issues.

[...]

PR #271, opened 2025-04-22
Migrate Layer Cake website to Svelte 5

rgieseke:
This migrates the website to Svelte 5, components remain as before.
So this is step 2 as discussed in https://github.com/mhkeller/layercake/issues/156#issuecomment-2722042331

(This should come after adding more tests from #261 and #270, i've run them only locally.)


rgieseke:
Sure, marking as draft makes sense!

src/routes/components/[slug]/+page.svelte
```diff
@@ -8,20 +8,15 @@
...
-	/** @type {import('./$types').PageData} */
-	export let data;
-	let { slug, content, active } = data;
-	$: slug = data.slug;
-	$: content = data.content;
-	$: active = data.active;
+	/** @type {import('./$types').PageProps} */
+	let { data } = $props();
+
+	let active = $derived(data.slug);
```
Comment (teonbrooks) [inline src/routes/components/[slug]/+page.svelte:19]:
```suggestion
	let active = $derived(data.active);
```

Comment (rgieseke) [reply src/routes/components/[slug]/+page.svelte:19]: Good catch, thanks!

mhkeller:
https://github.com/mhkeller/layercake/pull/261 and https://github.com/mhkeller/layercake/pull/270 are done but I think we still want to do https://github.com/mhkeller/layercake/pull/276 before tackling this one

mhkeller:
Noting that this PR and https://github.com/mhkeller/layercake/pull/281 will have to be merged into `main` together.

rgieseke:
There should probably be a test for the graphics on the landing page, too.

Not sure if you noticed already that the Beeswarm plot is not looking right. (When you click through it's alright.)

image

mhkeller:
There are a lot failing due to minor pixel differences. But there are others those where the reactivity is broken. Here's one example where I fixed it by cutting out intermediary derived values: https://github.com/mhkeller/layercake/pull/271/commits/51f556b4161dbb400aba980e75a30bab92bb0fe6

The migration script added in a temporary `run` function to some of the components so those are the ones I'm tackling first. 

Once the obviously broken ones are done, maybe we re-run the screenshots to fix the minor differences.

mhkeller:
> There should probably be a test for the graphics on the landing page, too.
> 
> Not sure if you noticed already that the Beeswarm plot is not looking right. (When you click through it's alright.)
> image

I didn't see your message before I posted that comment. Yep I think that falls under the category of "broken reactivity." Generally if the chart is bunched up into the top left corner, it means it rendered server-side but never got updated `$width` and `$height` values on the client.

mhkeller:
but yes, it would be a good idea to add a home page test. Probably best to add that as a new PR using the current `main` snapshots as a baseline

mhkeller:
Awesome. After removing `run` now it breaks in both places, so it's at least consistent. This one is definitely weird. One thing to be aware of is the migration command added snippets in some places but the main library doesn't support that yet. Ideally we could get all of the components working properly without rebuilding the main `LayerCake.svelte` component

mhkeller:
I updated a few of them but I'm running into an issue where

1. Canvas components and examples are running into an infinite loop, likely due to issues raised here: https://github.com/mhkeller/layercake/issues/250
2. So I tried commenting out the canvas examples and components but `npm run build` still fails because the prerender step is looking for those pages. So I added this in the svelte config

```js
prerender: {
  handleHttpError: 'warn'
}
```

Next step is to continue going through components and examples and getting them to match the screenshots.

mhkeller:
@rgieseke I tried manually running the update snapshots workflow but I get this error: https://github.com/mhkeller/layercake/actions/runs/16391856225/job/46318879345

mhkeller:
FYI the above commit now has the REPL links being created dynamically. It still needs some work standardizing the import paths since the REPL doesn't have any folders. But it will be a huge time save and ensure that the REPL links always match what's on the website. Unlike before, the URLs won't be stable, it's just encoding the state of the example in the URL hash.

mhkeller:
I was running into an error in the REPL that I think comes from putting `setContext` in a reactive statement. I removed that [here](https://github.com/mhkeller/layercake/pull/271/commits/b0e9efdd1e15328964da1e8338e2c435896e669a) since I don't think this call is necessary since the stores will be reactive and the context does not need to be reset. But I want to keep my eye on it to make sure this didn't cause any issues.

mhkeller:
@rgieseke I'm having a hard time seeing the artifacts for this failed test. Are you able to see what's wrong with it? I think it's on the new gallery image.

mhkeller:
I fixed it by checking out a new branch based on this one, running the action on that and cherry picking the commit. Not ideal but it works. This needs some testing / clicking around to make sure it's all good.

rgieseke:
Awesome! I've been checking this out but will give it a further test. 
I saw that a pnpm lock file was committed. I assume that was unintentional?

mhkeller:
Ya I find the darwin ones helpful for local development so I can see if something is wrong (or fixed) without having to push it and wait a few minutes. 

Totally fine for me to add the timeout if that fixes it – or does that difference still appear?

Issue #255, opened 2025-03-13
Convert examples and components on the website to runes


mhkeller:
The REPL examples also need to be updated.

mhkeller:
Closed via #271

PR #281, opened 2025-05-13
WIP: Svelte 5 migration components

rgieseke:
First steps to migrate the components to Svelte 5 syntax (#255).
Some are not working yet but wanted to put the first explorations somewhere.

This depends on the website being converted to Svelte 5 syntax (#271)  which this PR includes (making it a bit hard to read in its current state).

Once/before ready the following questions need to be considered

- [ ] update repl examples
- [ ] decide how to link to or include Svelte 3/4 compatible components (could be link, e.g. [GitHub tag](https://github.com/mhkeller/layercake/tree/25d13dd7ac1cbef2122b06d52ee00283507f5d19/src/_components))


mhkeller:
Ya I've been wondering how best to preserve the old documentation. In addition to the component and chart examples, I wonder if the guide, too, should be preserved. The easiest way could be to duplicate the repo and host it on gh-pages as http://mhkeller.github.io/layercake-prerunes or something. That's a bad URL name and so it should be improved. Maybe layerake-v8

I can put a robots.txt in the root to prevent search engines collecting it so that people don't accidentally end up on the old site. Could also add a prominent banner at the top

rgieseke:
> I've been wondering how best to preserve the old documentation. In addition to the component and chart examples, I wonder if the guide, too, should be preserved.

Maybe we could keep them in the same repo and website?
Copy the old components, guides and examples to a subdirectory, like `v8/components` or `legacy/components`? 
Maybe one could even have two versions of the Layer Cake library in the same repo, one for Svelte 5 and the current versions?

Could make sense to hide them from search engines with `robot.txt` or tags.

mhkeller:
My thinking in putting it in its own repo and essentially freezing it in time is that there will come a day, possibly Svelte 6 where `$:` syntax won't work. In that case, I'm not sure how to keep compiling the old examples. Or, if the site ever needs a rearchitecting, future people won't have to worry about this part of the site, which should be seeing diminishing viewership over time.

mhkeller:
It may be best just to combine this with https://github.com/mhkeller/layercake/pull/271, or otherwise rebase have that branch as the merge target. 

FYI I made an archived repo and website for the old site https://mhkeller.github.io/layercake.pre-runes

PR #282, opened 2025-05-13
Improve percentRange docstring in AnnotationsData component

rgieseke:
Just a tiny thing I spotted while working on Svelte 5 conversions.

The JSDoc string should be `pr` here I think:

```js
  /** @type {boolean} [percentRange=false] - If `true` will set the `top` and `left` CSS positions to percentages instead of pixels. */
  export let pr = $percentRange;

  $: units = pr === true ? '%' : 'px';
```

I'm not sure whether the `false` default is correct (probably indirectly from the `$percentRange` default value.


mhkeller:
Good catch!

rgieseke:
> Good catch!

All thanks to the tests :sunglasses:

PR #293, opened 2025-07-14
Use Playwright Docker image

rgieseke:
This PR adds using an official Playwright image to run the tests in.

I hope that this would help making tests more stable and potentially even locally reproducible.

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -8,6 +8,10 @@ jobs:
   test:
     timeout-minutes: 60
     runs-on: ubuntu-latest
+    container:
+      # https://playwright.dev/docs/docker
+      image: mcr.microsoft.com/playwright:v1.54.0-noble
+      options: --user 1001
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


rgieseke:
It still downloads a fresh (and probably more recent) copy of Chromium. It skips downloading the "Downloading FFMPEG playwright build v1011" stuff as this is already part of the Docker image I believe.

I came across this in this blogpost:

https://scottspence.com/posts/testing-with-vitest-browser-svelte-guide#the-playwright-container-advantage

The official docs are here:

https://playwright.dev/docs/docker

This seems to be used quite often:

https://github.com/search?q=mcr.microsoft.com%2Fplaywright++language%3Ayaml++path%3A%2F%5E%5C.github%5C%2Fworkflows%5C%2F%2F&type=code

For example in the Playwright repo itself: 

https://github.com/microsoft/playwright.dev/blob/24ff9f6da2796e588607e052b7a9e3e64616442a/.github/workflows/azure-static-web-apps-delightful-forest-0a29f6210.yml#L55

PR #294, opened 2025-07-14
Only run the doc build and deployment in the main repo

rgieseke:
This avoids build errors in forks which can't (and probably shouldn't) deploy the docs. 

It's hard to test but I used something similar in another project.


rgieseke:
Example error in my fork: https://github.com/rgieseke/layercake/actions/runs/16261274224

"Get Pages site failed. Please verify that the repository has Pages enabled and configured to build using GitHub Actions"

PR #297, opened 2025-07-15
Add workflow to update Playwright snapshots

rgieseke:
This adds a workflow to manually update snapshots on a PR branch after a desired change. This should ensure that the changes use the same Docker image/environment as tests later.

One needs to click in the Actions tab and select the branch.

image

You can see my test of the workflow here:

https://github.com/rgieseke/layercake/pull/5

Sorry again for the PRs I accidentally started in this repo while testing!

It would be nice to add some output (something for later maybe), but the PR comments were not working because the run is not directly attached to a PR.
Currently the changes are only viewable in the diff.

### DIFF
--- a/.github/workflows/update-snapshots.yml
+++ b/.github/workflows/update-snapshots.yml
@@ -1,0 +1,57 @@
+name: Update Playwright Snapshots
+
+on:
+  workflow_dispatch:
+
+jobs:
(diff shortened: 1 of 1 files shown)


mhkeller:
Awesome very useful! Thanks!

PR #300, opened 2025-07-22
Add first draft of development docs

rgieseke:
As suggested in https://github.com/mhkeller/layercake/pull/271#issuecomment-3092390012 I drafted some notes in a DEVELOPMENT.md file on how the project is set-up and the tests are to be run.

Could probably also go somewhere else in the docs and certainly will need to be extended but it's a first draft.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -1,0 +1,51 @@
+# Layer Cake Development
+
+## Project Structure
+
+The main `LayerCake` library which you get when you
+
+```js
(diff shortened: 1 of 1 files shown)


mhkeller:
Awesome thank you!

One thing that I think would be good to alter about the tests would be instead of skipping ones that don't have `$props()` and therefore no table, they should be tested to make sure they don't have any props. I think now if you just deleted the `$props()`, it wouldn't cause the test to fail.

PR #320, opened 2025-08-19
Add timeout for clipboard test

mhkeller:
After https://github.com/mhkeller/layercake/pull/313 I was getting a failed test intermittently for the AxisRadial and Bar clipboard tests, this seems to fix it.


rgieseke:
Hm, odd, but it works for now in CI ...

PR #325, opened 2025-08-19
Fix error with too many args and simplify arrowUtils

rgieseke:
Error was

```
/layercake/src/_modules/arrowUtils.js:62:33
Error: Expected 1-2 arguments, but got 3. 
            data = data.map((d, i) => {
                        return [xValue.call(data, d, i), yValue.call(data, d, i)];
                });

/layercake/src/_modules/arrowUtils.js:62:58
Error: Expected 1-2 arguments, but got 3. 
            data = data.map((d, i) => {
                        return [xValue.call(data, d, i), yValue.call(data, d, i)];
                });
```

I simplified the expression to directly call `xValue`/`yValue`.

### DIFF
--- a/src/_modules/arrowUtils.js
+++ b/src/_modules/arrowUtils.js
@@ -58,8 +58,8 @@ export function swoopyArrow() {
 	}
 
 	function render(data) {
-		data = data.map((d, i) => {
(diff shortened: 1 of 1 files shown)


rgieseke:
Yea, Copilot had something that it might be useful if the `this` context would change or something, maybe it came from such a use-case.

Issue #329, opened 2025-08-23
TS error: not detecting that children are allowed


mhkeller:
You can see this in the [layercake-template repo](https://github.com/mhkeller/layercake-template) if you hover over the `Svg` layout component:

<img width="749" height="516" alt="Image" src="https://github.com/user-attachments/assets/fb390ccf-fe33-4ed1-90a8-aef71a6a81ac" />

It looks like it comes from the type definition where `  [x: string]: never;` instead of `() => any`. These types are auto-generated so I'm not sure what the right fix is here. I tried making `children` required but that didn't change the auto-generated types file.

```svelte
/** @typedef {typeof __propDef.props}  SvgProps */
/** @typedef {typeof __propDef.events}  SvgEvents */
/** @typedef {typeof __propDef.slots}  SvgSlots */
/** SVG layout component */
export default class Svg extends SvelteComponentTyped<{
    [x: string]: never;
}, {
    [evt: string]: CustomEvent<any>;
}, {}> {
}
export type SvgProps = typeof __propDef.props;
export type SvgEvents = typeof __propDef.events;
export type SvgSlots = typeof __propDef.slots;
import { SvelteComponentTyped } from "svelte";
declare const __propDef: {
    props: {
        [x: string]: never;
    };
    events: {
        [evt: string]: CustomEvent<any>;
    };
    slots: {};
};
export {};

```

rgieseke:
If I run `npx sv migrate svelte-5` on the repo and fix some things (e.g. rename `derived` to let it run) I can get something like the following, that is, it uses the new `Component` type and the error in the template disappears:

cat dist/layouts/Svg.svelte.d.ts

```
export default Svg;
type Svg = {
    $on?(type: string, callback: (e: any) => void): () => void;
    $set?(props: Partial<Props>): void;
};
/** SVG layout component */
declare const Svg: import("svelte").Component<{
    /**
     * The layout's `<svg>` element. A useful prop to bind to.
     */
    element?: SVGElement | undefined;
    /**
     * Set the layout's z-index.
     */
    zIndex?: number | undefined;
    /**
...
```

mhkeller:
Thanks for the research! So we should really just tackle migrating `LayerCake.svelte`. I had thought (maybe a year ago or so) we would have to migrate to state and get rid of the stores architecture but I don't think that's the case, which makes the migration easier.

rgieseke:
[...]

PR #331, opened 2025-08-24
Convert LayerCake.svelte lib to Svelte 5

rgieseke:
As discussed in #329


rgieseke:
On `build` I get

```
src/lib/LayerCake.svelte:204:28 This reference only captures the initial value of `yReverseValue`. Did you mean to reference it inside a closure instead?
https://svelte.dev/e/state_referenced_locally
202:   const _rNice = writable(rNice);
203:   const _xReverse = writable(xReverse);
204:   const _yReverse = writable(yReverseValue);
                                               ^
205:   const _zReverse = writable(zReverse);
206:   const _rReverse = writable(rReverse);
```

mhkeller:
I've never liked how the LayerCake component has all of this code for x, y, z and r even if the user hasn't defined those. It would be nice if things were more dynamic where it looked at whether those accessors were set and then created properties as a result. I'm not sure if this will work with Svelte's reactivity but just putting it out there.

mhkeller:
> Maybe we can handle yReverseValue differently and then just use the derived store.

Ya that makes sense.

mhkeller:
We should pick this back up. If we keep stores, then we probably can't have it be more dynamic but I'd have to look into it. I had originally thought stores were going away but I think they aren't? If they remain, then can keep our destructured props from the layercake context. There is some new stuff around [createContext](https://svelte.dev/docs/svelte/context) that maybe would be helpful such as adding type hints?

rgieseke:
Yes, the `createContext` seems quite useful!

rgieseke:
Here is an attempt to convert the `LayerCake` component without stores to Svelte 5.

It seems to be the recommended way, from https://svelte.dev/docs/kit/state-management#Using-state-and-stores-with-context

> You also use stores from svelte/store for this, but when using Svelte 5 it is recommended to make use of universal reactivity instead.

I think it makes the `LayerCake` component simpler to understand and for a user it might be easier to understand a LayerCake component without the 'magic' that comes from importing a store from a context and then using it pre-fixed with a `$`.

The basic approach I used is to store original prop values or derived variables (`_d`) in an object with getters so that they can be accessed in child components.

The import in a child component is then

```js
const { data, xGet, yGet, xScale, yScale } = $derived(getLayerCakeContext());
 ```

for using them as `data` instead of `$data`.

Alternatively, one could them import them as `context`, `c`, or `chart` 

```js
const context = getLayerCakeContext()
```

for using them as `context.data`.

It would be a breaking change but updating user's components could likely be done with search and replace or a migration script.

rgieseke:
For the shared context I used a `context.ts` file which seemed easier to write but it of course has lots of room for improvements. I think internally it's probably fine to use TypeScript? I saw you used `types.d.ts` in https://github.com/mhkeller/layercake-annotations - did that work well for you?

I think it would be nice to keep the downloadable components with JSDoc so that they can be dropped in any project.

mhkeller:
Thanks for making more progress on this! I agree that I think moving away from `$` will make it easier for users. The `$derived(getLayerCakeContext())` is a bit ugly. Maybe in the docs we use `const c = getLayerCakeContext()` and the usage would be `c.width`

Compared to the original syntax where you had to add a `$`, it's not that much more typing.

The types in that project seemed to work fine? I don't fully remember. Lemme dig into the code a bit! (I converted this to draft so as to not trigger more CI runs but I think it might have to be merged with main to actually get that fix...)

rgieseke:
Agreed, I've updated the syntax to use `const c = getLayerCakeContext()` and `c.width` etc.

I think the approach with using declaration files as you did in `layercake-annotations` can be quite useful, could be helpful with re-using definitions as discussed in #298.

mhkeller:
Superseded by #422, which contains every commit from this branch.

PR #335, opened 2025-08-27
Use rollup plugin-dsv to typecast CSV data

rgieseke:
Following up on https://github.com/mhkeller/layercake-template/pull/43 and some earlier attempts of mine to fix some CSV import related type issues I propose to type the CSV data as:

```js
module '*.csv' {
    const data: { [key: string]: any }[];
    export default data;
}
```
and to convert the types to `string`  or `number` using `processRow` of the rollup [plugin-dsv](https://www.npmjs.com/package/@rollup/plugin-dsv):

```js
dsv({
	processRow: row => {
		Object.keys(row).forEach(key => {
			var value = row[key];
			if (value !== '' && !isNaN(value)) {
				row[key] = Number(value);
			} else {
				row[key] = value;
			}
		});
	}
})
```
I had tried using `string|number` instead of `any` in the type declaration but that lead to problems with the `timestring` examples where the field is changed to be of type `Date`.

It changes years like '1979' in the examples to be of type number instead of string because these are now already converted to `number` in the import.

CSV type definition are moved to `app.d.ts` from `global.d.ts`, this seems to be the default now, couldn't really find when or where this was changed, but `app.d.ts` seems to be the default now (https://svelte.dev/docs/kit/types#app.d.ts)

The `data.forEach` typecasting from the examples is removed, only the `Date` conversions remain. I think this makes it a little bit easier to follow the examples. One could argue that now it's a bit hidden in the config file, but if the data was given as JSON it would have the types already and no conversion would be necessary.  

A lot of errors in `svelte-check` like the following are thus removed.
```
Error: Parameter 'd' implicitly has an 'any' type. (js)
	data.forEach(d => {
		xKey.forEach(name => {
```

Errors are reduced from 549 in 141 files to 468 in 106 files.


rgieseke:
For context, these are the example CSV files in the repo (first 3 lines):

```
==> src/_data/cars-2.csv <==
Name,Miles_per_Gallon,Cylinders,Displacement,Horsepower,Weight_in_lbs,Acceleration,Year,Origin
chevrolet chevelle malibu,18,8,307,130,3504,12,1970,USA
buick skylark 320,15,8,350,165,3693,11.5,1970,USA

==> src/_data/dates-april.csv <==
timestring
2018-04-29T02:22:38Z
2018-04-29T02:19:25Z

==> src/_data/dates.csv <==
timestring
2018-07-22T22:25:55Z
2018-07-22T19:35:29Z

==> src/_data/days.csv <==
timestring
2018-07-22T22:25:55Z
2018-07-22T19:35:29Z

==> src/_data/familyTree.csv <==
name,parent
Eve,
Cain,Eve

==> src/_data/fruit.csv <==
month,apples,bananas,cherries,dates
2015-04-01,3840,1920,960,400
2015-03-01,1600,1440,960,400

==> src/_data/fruitGroups.csv <==
fruit,value
apples,3240
bananas,1920

==> src/_data/fruitOrdinal.csv <==
year,apples,bananas,cherries,dates
2019,3840,1920,960,400
2018,1600,1440,960,400

==> src/_data/groups.csv <==
year,value
1979,2
1980,3

==> src/_data/points.csv <==
myX,myY
1979,7.19
1980,7.83

==> src/_data/pointsFour.csv <==
myX,myY
1979,10.27
1980,1.99

==> src/_data/pointsThree.csv <==
myX,myY
1979,5.43
1980,7.27

==> src/_data/pointsTwo.csv <==
myX,myY
1979,5.03
1980,3.99

==> src/_data/radarScores.csv <==
name,fastball,change,slider,cutter,curve
Allison,10,0,4,8,5

==> src/_data/us-senate.csv <==
state_name,party,name,gender,date_of_birth
Alaska,R,Daniel Sullivan,male,1964-11-13
Alaska,R,Lisa Murkowski,female,1957-05-22
```

src/routes/_examples/ColumnStacked.svelte
```diff
@@ -21,11 +21,6 @@
 	/* --------------------------------------------
 	 * Cast data
 	 */
```
mhkeller, line 23: Should get rid of this comment too

mhkeller:
It's nice to get rid of the casting to make the examples a little easier to follow. Because it does make things a little hidden, let's change the comment `This example loads csv data as json using @rollup/plugin-dsv` to `This example loads csv data as json and converts numeric columns to numbers using @rollup/plugin-dsv. See vite.config.js for details`

We would also have to update [layercake-template](https://github.com/mhkeller/layercake-template) so that the examples download propertly

src/routes/_examples/ColumnStacked.svelte
rgieseke: Good catch, done!

rgieseke:
Great, I've updated the comment and will create a PR for the template repo.

mhkeller:
Looks good thank you! I had though we could also test for booleans and null because `Number(null) === 0` but since these values will always be strings. I think it's alright.

rgieseke:
> @rgieseke The did a quick test on the column example and that one looks like it still has the number casting in the example. That should be removed, right? 

Which file(s)? I don't spot it right now.

> And the values in the arrow config changed to numbers.

Yes, since the 'numbers' come in as numbers by default now it I changed that.

mhkeller:
> Which file(s)? I don't spot it right now.

My pull must have been weird. A fresh pull now has it fixed!

PR #394, opened 2026-04-13
Pin npm

rgieseke:
Sorry, meant to run this on my fork :see_no_evil:


mhkeller:
What if we switched to pnpm?

rgieseke:
Sorry again for this accidental PR, it seems removing the pin to 11 worked (see #393).

I used pnpm in some projects and it looks to be used in Sveltekit as well, so i'm fine with either.

Issue #50, opened 2021-11-17
Two plots on one <Canvas> causes update loop?


abalmos:
Thank you for writing LayerCake, it is a fantastic base to make custom plots on top of! We are moving from a more "component" based library and really enjoy the control Layer Cake provides. Particularly being able to trivally combine SVG (ease) with Canvas (high volume data layers).

That said, I have been playing with Canvas based plots today in anticipation of needing to plot a large amount of data soon. It seems that putting two components in one `<Canvas>` results in an update loop that never ends.

I wonder if I am doing something wrong?

You can see the issue [in this REPL](https://svelte.dev/repl/1f88e240cc9f46b5bdee415a7658fffd?version=3.44.1). It works as shown with two `<Canvas>` tags. You can see the issue by moving `<Scatter />` into `<Line />`'s `<Canvas>` (while also commenting the two scaleLayer and clear rect lines from `<Scatter />`). WARNING: It will bind up your tab as soon as the REPL rebuilds.

Also, I wonder if multiple `<Canvas>` is better in the end anyway?

mhkeller:
Hey @abalmos, sorry I didn't see this earlier. Hm that is interesting and I hadn't seen that before. The issue seems to be that Svelte believes the `$ctx` variable is modified on lines 22 and 24 and `Scatter.svelte` when the `fillStyle` and `strokeStyle` are being set. Check out [this example](https://svelte.dev/repl/9f0ab6130a7f4566a2d41e51e48892a1?version=3.44.1) that works in one `<Canvas>` tag without those lines. 

I'm not sure if there's a workaround since this seems to be just how Svelte's reactivity works but maybe @rich-harris has some other ideas. It would be nice if you didn't have to add multiple canvas tags for every layer although that seems like a fine workaround.

mhkeller:
I haven't tested this but one idea is to create a wrapper that loads the two components from within the `:if ($ctx)` block but not sure if that will solve it.

mhkeller:
This looks like a svelte bug and hopefully will be fixed soon.

abalmos:
@mhkeller thanks for checking into this. My attention was stolen away from the project, but I will return soon. Is there a Svelte issue to track?

mhkeller:
It may be this one: https://github.com/sveltejs/svelte/issues/4933

mhkeller:
FYI I added this to the docs: https://github.com/mhkeller/layercake/commit/e459efff16f627507c80364de53d0b22a79884b0

PR #424, opened 2026-08-23
Support multiple layers per <Canvas>

### LINKED ISSUE #250: Canvas & Svelte 5 infinite effect loops

mhkeller:
`<Canvas>` now owns its canvas. The element covers the whole chart container and the layout scales, clears and moves the origin before every repaint. Much inspired by [@techniq](https://github.com/techniq)'s work in LayerChart and [this ancient conversation](https://github.com/techniq/layerchart/issues/183). Components stop managing the canvas and hand over a draw function instead:

```js
const k = getLayerCakeContext();
const canvas = getCanvasContext();

canvas.draw(ctx => {
	for (const d of k.data) {
		ctx.beginPath();
		ctx.arc(k.xGet(d), k.yGet(d), 5, 0, 2 * Math.PI);
		ctx.fill();
	}
});
```

Whatever the function reads (props, `$state`, `k.*`) triggers a repaint. Several components can share one `<Canvas>` and paint in order, so the old one-`<Canvas>`-per-component workaround (#50) is gone. Since the origin moves by the padding, canvas drawings can reach into the margins the way Svg and Html children always could.

## API

- `getCanvasContext()` returns a typed context: `draw(fn)`, `redraw()` for things Svelte can't track and a read-only `ctx`
- `k.pointer(event)` gives chart-area `[x, y]` on any layer. Layers cover different boxes now, so `offsetX` shifts meaning depending on where you listen. `k.pointer` doesn't
- `overflow` on `<Canvas>`, the same prop as the other layouts. `"hidden"` clips at the chart area
- `draw()` must be called during component setup and throws a clear error anywhere else
- a one-time console warning when something else resizes the canvas (a v10-style `scaleCanvas` call, usually)

## Also in here

- the WebGL element no longer spills past the container by the padding. Its CSS was over-constrained
- main's idempotent `scaleCanvas` and its test, brought forward so the merge is clean
- the paint pass lives in `helpers/paintLayers.js` with unit tests for order, clipping and error isolation
- the three canvas docs components rewritten. Guide and CHANGELOG updated
- MapLayered puts two components on one canvas: a dot for each state too small to label
- e2e checks: the canvas covers the container, canvas points land on the Svg points, layers paint in order

## Worth knowing

- A layer remounted by `{#if}` goes to the top of the stack. An `order` option can come later but could be clunky.
- The full WebGL version of this pattern needs its own design (no `translate` in GL) and stays on the roadmap


mhkeller:
Merged into main via #429. GitHub's stack merge kept refusing this PR after #422 was closed, so #429 is the same rebased branch opened directly against main.

